PR #2, opened 2018-08-01
Migrate from turtles to droste

pepegar:
Use droste for recursion schemes


src/main/scala/freestyle/Optimize.scala
```diff
@@ -0,0 +1,64 @@
...
+   *
+   * {{{
+   * case class Product(field1: String, field2: OtherField)
+   * }}}
+   */
+  def nestedNamedTypesTrans[T](implicit T: Basis[Schema, T]): Trans[Schema, Schema, T] = Trans {
+    case TProduct(name, fields) =>
+      TProduct[T](
+        name,
+        fields.map { f: Field[T] =>
```
diesalbla: Same as  the comment for the `optimize.md` file.

src/main/scala/freestyle/schema.scala
```diff
@@ -62,4 +63,35 @@ object Schema {
...
+    case TNamedType(name) => name
+    case TOption(value)   => s"Option[$value]"
+    case TMap(value)      => s"Map[String, $value]"
+    case TList(value)     => s"List[$value]"
+    case TRequired(value) => value
+    case TCoproduct(invariants) =>
+      invariants.toList.mkString("Cop[", " :: ", ":: TNil]")
+    case TSum(name, fields) =>
+      val printFields = fields.map(f => s"case object $f extends $name").mkString("\n  ")
+      s"""
```
diesalbla: Another use for `stripMargin`.

src/main/scala/freestyle/schema.scala
```diff
@@ -62,4 +63,35 @@ object Schema {
...
+sealed trait $name
+object $name {
+  $printFields
+}
+"""
+    case TProduct(name, fields) =>
+      val printFields = fields.map(f => s"${f.name}: ${f.tpe}").mkString(", ")
+      s"""
```
diesalbla: Would the next three lines be the same as the following line? 
```Scala
"@message case class $name($printFields)" 
```

src/main/scala/protobuf/schema.scala
```diff
@@ -79,4 +80,61 @@ object Schema {
...
+  $printAliases
+}
+"""
+    case TMessage(name, fields, reserved) =>
+      val printReserved = reserved.map(l => s"reserved " + l.mkString(", ")).mkString("\n  ")
+      def printOptions(options: List[Option]) =
+        if (options.isEmpty) {
+          ""
+        } else {
+          options.map(printOption).mkString(" [", ", ", "]")
```
diesalbla: In the `mkString`, I had some problem reading apart each quote and comma. It may be worth using named parameters: 
```Scala
mkString( start = " [" , sep = ", " , end = "]" ) 
```

docs/src/main/tut/optimizations.md
```diff
@@ -42,32 +43,39 @@ case class Product(field1: String, field2: OtherField)
...
 and then apply the `namedTypes` combinator to the AST:
 
 ```tut:invisible
-def ast[T](implicit T: Birecursive.Aux[T, Schema]): T = TNull[T]().embed
+def ast = Fix(TNull[Fix[Schema]]())
 ```
 
 ```tut
-ast[Mu[Schema]].transCataT(namedTypes)
+val optimization = Optimize.namedTypes[Fix[Schema]]
```
andyscott: You should be able to use `Mu` still, if you want.

src/main/scala/avro/Protocol.scala
```diff
@@ -0,0 +1,53 @@
...
+    messages: List[Protocol.Message[A]]
+)
+
+object Protocol {
+  import Schema._
+
+  case class Message[A](name: String, request: A, response: A)
+
+  def fromProto[T](proto: AvroProtocol)(implicit T: Embed[Schema, T]): Protocol[T] = {
+    val toFreestyle: AvroSchema => T = scheme.ana(fromAvro)
```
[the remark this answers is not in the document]
pepegar: Ouch, you're right! I think I copypasted :)

src/main/scala/freestyle/schema.scala
pepegar: Indeed, good catch! ;)

src/main/scala/protobuf/schema.scala
pepegar: :+1:

docs/src/main/tut/optimizations.md
pepegar: Yeah, IDK why I didn't see it in droste :) Thanks!

PR #10, opened 2018-08-03
Swagger schema

pepegar:
This PR:

- implements openapi schema (including the jsonschema subset) from https://swagger.io/specification
- renames current schemas (avro, protobuf, & frees)
- creates a `droste.contrib` file with stuff that potentially can be backported to droste
- creates JSON encoders for openapi representations


src/main/scala/freestyle/Transform.scala
```diff
@@ -17,58 +17,58 @@
...
+    case AvroF.TArray(item)     => TList(item)
+    case AvroF.TMap(values)     => TMap(values)
+    case AvroF.TRecord(name, _, _, _, fields) =>
       TProduct(name, fields.map(f => Field(f.name, f.tpe)))
-    case AvroSchema.TEnum(name, _, _, _, symbols) => TSum(name, symbols)
-    case AvroSchema.TUnion(options)               => TCoproduct(options)
-    case AvroSchema.TFixed(_, _, _, _) =>
+    case AvroF.TEnum(name, _, _, _, symbols) => TSum(name, symbols)
+    case AvroF.TUnion(options)               => TCoproduct(options)
+    case AvroF.TFixed(_, _, _, _) =>
```
diesalbla: The documentation on the `fixed` in the Avro docs is sparse. However, if it is intended to represent byte arrays, then [scodec](http://scodec.org/) offers support for that. On the other hand, the [refined](https://github.com/fthomas/refined) allows restricting the length of a string. However, I would be wary of bringing too many libraries. 

In any case, if we do not support them, we should add a more explicit message or exception, saying that `skeuomorph` library does not have support for Avro fixed types.

pepegar: My initial idea was to drop Fixed from our ADT, but using refined seems interesting. I kinda _discard_ scodec because one can't use fixed length BitVectors/ByteVectors there, right?

I'll create an issue to remind me to implement fixed usign refined using `Array[Byte] Refined MaxSize[N]`

PR #13, opened 2018-10-10
Frees & Avro schema fixes

pepegar:
This PR:

- renames `frees/service.scala` to `frees/protocol.scala` to be homogeneous with avro & protobuf
- deletes `droste` contrib module that has been already contributed to droste.
- adds helper methods to avro ADT
- creates transformer between `AvroF` & `FreesF` schemata
- creates a rendering algebra from `AvroF[A]` to `circe.Json` and creates the renderer from `avro.Protocol[A]` to `circe.Json`.


src/main/scala/avro/schema.scala
```diff
@@ -142,18 +188,58 @@ object AvroF {
...
+          fields.map(
+            f =>
+              Json.obj(
+                "name"    -> Json.fromString(f.name),
+                "aliases" -> Json.arr(f.aliases.map(Json.fromString): _*),
+                "type"    -> f.tpe
+            )): _*)
+      )
+      val withNamespace = namespace.fold(base) { n =>
+        base deepMerge Json.obj("namespace" -> Json.fromString(n))
```
diesalbla: In this particular case, would `deepMerge` be the same as: 
```Scala
base.mapObject( _.add ("namespace", Json.fromString(n) ) )
```

src/main/scala/avro/schema.scala
```diff
@@ -142,18 +188,58 @@ object AvroF {
...
+        "type"   -> Json.fromString("map"),
+        "values" -> values
+      )
+    case TRecord(name, namespace, aliases, doc, fields) =>
+      val base: Json = Json.obj(
+        "type" -> Json.fromString("record"),
+        "name" -> Json.fromString(name),
+        "fields" -> Json.arr(
+          fields.map(
+            f =>
```
diesalbla: This 4-line lambda should go to an auxiliary function `def fieldToObj`?

src/main/scala/avro/schema.scala
```diff
@@ -142,18 +188,58 @@ object AvroF {
...
+      Json.obj(
+        "type"  -> Json.fromString("array"),
+        "items" -> item
+      )
+    case TMap(values) =>
+      Json.obj(
+        "type"   -> Json.fromString("map"),
+        "values" -> values
+      )
+    case TRecord(name, namespace, aliases, doc, fields) =>
```
diesalbla: If you are into a bit of Origami-like programming, you could write this case as follows: 

- From each of `namespace`, `aliases`, and `doc`, compute an `Option[( String, Json) ]`: 

```Scala
val namespaceOF = namespace.map( n => "namespace" -> Json.fromString(n) _" 
val aliasesOF  = if (aliases.isEmpty) None else Some( "aliases" -> Json.arr(aliases.map(Json.fromString): _*) )
val docOF  = doc.map( f => "doc" -> Json.fromString(f) )
```
- Add a function `addField(js: Json, of: Option[(String, Json)])`, which adds a field to a Json object. 
- Get the result as `List(namespaceOF,  aliasesOF, docOF).foldLeft(base)(addField)`

An alternative, but I would not say if it is cleaner. At your taste.

diesalbla: Another alternative is that you keep the list of fields, with `type`, `name`, `fields`; and you make `namespaceOf` or `aliasesOf` into lists (empty or single), and you don't apply the `Json.obj` constructor until the end.

build.sbt
```diff
@@ -59,7 +59,8 @@ lazy val commonSettings = Seq(
   ThisBuild / scalacOptions -= "-Xplugin-require:macroparadise",
   libraryDependencies ++= Seq(
     %%("cats-core"),
-    "io.higherkindness" %% "droste-core" % "0.3.1",
+    "io.higherkindness" %% "droste-core"     % "0.5.0",
+    "io.higherkindness" %% "droste-macros"     % "0.5.0",
```
juanpedromoreno: Odd formatting here, we might want to apply the scalafmt syntax also for the sbt files.. makes sense?

src/main/scala/freestyle/protocol.scala
```diff
@@ -0,0 +1,106 @@
...
+import qq.droste._
+
+sealed trait SerializationType extends Product with Serializable
+object SerializationType {
+  case object Protobuf       extends SerializationType
+  case object Avro           extends SerializationType
+  case object AvroWithSchema extends SerializationType
+}
+
+case class Protocol[T](
```
juanpedromoreno: final?

src/main/scala/freestyle/protocol.scala
```diff
@@ -0,0 +1,106 @@
...
+    |
+    |$printDeclarations
+    |$printServices
+    |
+    |}
+    """.stripMargin
+  }
+}
+
+case class Service[T](name: String, serializationType: SerializationType, operations: List[Service.Operation[T]])
```
juanpedromoreno: final case class?

src/main/scala/avro/Protocol.scala
```diff
@@ -35,6 +40,18 @@ object Protocol {
 
   case class Message[A](name: String, request: A, response: A)
 
+  object Message {
+    def toJson[T](message: Message[T])(implicit T: Project[AvroF, T]): Json = Json.obj(
+      "request" -> Json.arr(
+        Json.obj(
+          "name" -> Json.fromString("arg"), //TODO: is this doable?
+          "type" -> scheme.cata(AvroF.toJson).apply(message.request)
```
andyscott: I'd recommend caching/saving `scheme.cata(AvroF.toJson)` to a val (or lazy val) and then calling it here.

src/main/scala/avro/Protocol.scala
```diff
@@ -35,6 +40,18 @@ object Protocol {
...
+        )
+      ),
+      "response" -> scheme.cata(AvroF.toJson).apply(message.response)
```
andyscott: Same as above.

src/main/scala/avro/Protocol.scala
```diff
@@ -48,4 +65,58 @@ object Protocol {
...
+    case FreesF.TInt()                 => tInt()
+    case FreesF.TLong()                => tLong()
+    case FreesF.TBoolean()             => tBoolean()
+    case FreesF.TString()              => tString()
+    case FreesF.TByteArray()           => tBytes()
+    case FreesF.TNamedType(name)       => tNamedType(name)
+    case FreesF.TOption(value)         => tUnion(NonEmptyList(tNull[T]().embed, List(value)))
+    case FreesF.TList(value)           => tArray(value)
+    case FreesF.TMap(value)            => tMap(value)
+    case FreesF.TGeneric(_, _)         => ??? // WAT
```
andyscott: For the long term, you could do a `TransM` through `Option` to account for this failure scenario.

src/main/scala/freestyle/protocol.scala
```diff
@@ -0,0 +1,106 @@
...
+object Service {
+  case class Operation[T](name: String, request: T, response: T)
```
andyscott: `final` :)

src/main/scala/freestyle/schema.scala
```diff
@@ -40,46 +40,49 @@ object FreesF {
   case class TOption[A](value: A)                              extends FreesF[A]
   case class TList[A](value: A)                                extends FreesF[A]
   case class TMap[A](value: A)                                 extends FreesF[A]
+  case class TGeneric[A](generic: A, params: List[A])          extends FreesF[A]
   case class TRequired[A](value: A)                            extends FreesF[A]
   case class TCoproduct[A](invariants: NonEmptyList[A])        extends FreesF[A]
   case class TSum[A](name: String, fields: List[String])       extends FreesF[A]
   case class TProduct[A](name: String, fields: List[Field[A]]) extends FreesF[A]
```
andyscott: even more `final` :)

src/main/scala/avro/Protocol.scala
pepegar: yeah, it's in my plans, but I need to do some changes to droste's traverse derivation, since its a bit far from flawless so far... (I need traverse to be able to use `M` version of recursion schemes, and I don't want to implement by hand now :))

PR #14, opened 2018-10-11
create a new approach for pretty printing

pepegar:
To be merged after #13 .

This PR investigates a new approach for pretty printing, using contravariant functors, and the `Decidable` & `Divisible` typeclasses (duals of `Alternative` & `Applicative` respectively).

It's basically a more principled way or pretty printing. Instead of creating strings and interpolating them manually, we create small `Printer`s (dual of `Parser`, if you will), and compose them with `Divisible`, `Decidable` & `Contravariant` operations.


src/main/scala/freestyle/print.scala
```diff
@@ -0,0 +1,141 @@
...
+  ): ((SerializationType, String), List[Service.Operation[T]]) =
+    s match {
+      case Service(name, serType, ops) =>
+        ((serType, name), ops)
+    }
+
+  /**
+   * needed to use SerializationType as a [[catz.contrib.Decidable]].
+   */
+  def serTypeEither(serType: SerializationType): Either[Either[Protobuf.type, Avro.type], AvroWithSchema.type] =
```
juanpedromoreno: Maybe we could use any other kind of coproduct type to make it simpler, shapeless coproduct perhaps?

pepegar: The idea of using either is because of Decidable, which uses Eithers directly. I'd prefer to keep Either to be able to contribute Divisible and Decidable to cats at some point

PR #17, opened 2018-10-20
New approach for pretty printers

pepegar:
clone of #14 . I don't really know when I closed it :)


src/main/scala/Printer.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  def newLine = konst("\n")
+
+  val string: Printer[String] = Printer(identity)
+
+  def optional[A](p: Printer[A]): Printer[Option[A]] =
+    Printer(_.fold("")(p.print))
+
+  def mkList[A](p: Printer[A], sep: String): Printer[List[A]] =
+    Printer(_ match {
```
juanpedromoreno:
```suggestion
    Printer({
```

src/main/scala/freestyle/print.scala
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package skeuomorph
+package freestyle
```
juanpedromoreno: We might want to create some tickets to rename this and also moving this project to the hk organization (under higherkindness.io group id). Makes sense?

src/main/scala/freestyle/print.scala
```diff
@@ -0,0 +1,141 @@
...
+        val printFields = fields.map(f => s"case object $f extends $name").mkString("\n  ")
+        s"""
+      |sealed trait $name
+      |object $name {
+      |  $printFields
+      |}
+      """.stripMargin
+      case TProduct(name, fields) =>
+        val printFields = fields.map(f => s"${f.name}: ${f.tpe}").mkString(", ")
+        s"@message case class $name($printFields)"
```
juanpedromoreno: final case class?

src/main/scala/freestyle/print.scala
```diff
@@ -0,0 +1,141 @@
...
+  def service[T](implicit T: Basis[FreesF, T]): Printer[Service[T]] =
+    ((konst("@service(") *< serializationType >* konst(") trait ")) >*<
+      (string >* konst("[F[_]] {") >* newLine) >*<
+      (mkList(operation, "\n") >* newLine >* konst("}"))).contramap(serviceTuple)
+
+  def option: Printer[(String, String)] =
+    (konst("@option(name = ") *< string) >*< (konst(", value = ") *< string >* konst(")"))
+
+  def proto[T](implicit T: Basis[FreesF, T]): Printer[Protocol[T]] = {
+    ((konst("pakage ") *< optional(string) >* (newLine >* newLine)) >*<
```
juanpedromoreno:
```suggestion
    ((konst("package ") *< optional(string) >* (newLine >* newLine)) >*<
```

src/main/scala/freestyle/print.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  def operation[T](implicit T: Basis[FreesF, T]): Printer[Service.Operation[T]] =
+    ((konst("def ") *< string) >*< (konst("(req: ") *< Printer(namedTypes[T] >>> schema.print)) >*< (konst("): ") *< Printer(
+      namedTypes[T] >>> schema.print)))
+      .contramap(opTuple)
+
+  def service[T](implicit T: Basis[FreesF, T]): Printer[Service[T]] =
+    ((konst("@service(") *< serializationType >* konst(") trait ")) >*<
+      (string >* konst("[F[_]] {") >* newLine) >*<
+      (mkList(operation, "\n") >* newLine >* konst("}"))).contramap(serviceTuple)
```
juanpedromoreno: Can we use https://github.com/higherkindness/skeuomorph/pull/17/files#diff-3dd9d92ac4124b698fac391a3f4df6aaR32 somehow?

juanpedromoreno: Or at least, I'd suggest for extracting that out to a `val`

src/main/scala/freestyle/print.scala
pepegar: Yep, I'll tackle those in #18

src/main/scala/catz/contrib/Divisible.scala
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package skeuomorph
+package catz
+package contrib
+
+import cats.Contravariant
+
+trait Divisible[F[_]] extends Contravariant[F] {
```
diesalbla: Is this related to [Ed Kmett's Divisible library](https://hackage.haskell.org/package/contravariant-1.4/docs/Data-Functor-Contravariant-Divisible.html)?

If it is, I think a similar thing is already [added to `cats`](https://github.com/typelevel/cats/pull/2034/files#diff-04759b52b88a7257bb687517dc29184dR12), under the inconspicuous name of `ContravariantMonoidal`.

pepegar: Yep, it's a port of kmett's Divisible indeed.  But `ContravariantMonoidal` is lacking `divide` right?

src/main/scala/Printer.scala
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+
+package skeuomorph
+
+import catz.contrib.Decidable
+import cats.syntax.compose._
+import cats.instances.function._
+
+case class Printer[A](print: A => String)
```
diesalbla: Was there a reason for using a `case class`?  Libraries like `circe` usually write the encoders or printers as traits with the abstract method: 
```Scala
trait Printer[A]{
    def print(a: A): String

    def contraMap[B](f: B => A): Printer[B]
}
```
There are some benefits of these approach. First, you save a bit of efficiency, but turning two objects (case class and functions) into a single one. Also, in Scala 2.12, thanks to the SAM syntax, you can assign a lambda to it 
```Scala
val print: Printer[Person] = { x => s"${x.name}, ${x.age} years." } 
```

On another matter: should we make `A` contravariant, `trait Printer[- A]`?

src/main/scala/catz/contrib/Divisible.scala
diesalbla: It is not _lacking_ it. It just so happens that, in a very curious leap of abstraction, they use `Semigroupal[F].product` to represent both the `divide` method from the `Divisible` type-class, and the `product` (the old tie fighter `|@|`) from the `Applicative` class.

diesalbla: For the same token, the method `conquer` has a very similar signature to `ContravariantMonoidal.trivial`. Nevertheless, even if you base the code of `cats`, It may be easier to keep the Divisible syntax.

pepegar: Yep, I'm keeping the syntax but dropping the typeclass in favor of `ContravariantMonoidal`. Thanks for pointing me to it Diego!

src/main/scala/Printer.scala
```diff
@@ -0,0 +1,66 @@
...
+  def contramap[B](f: B => A): Printer[B] = Printer(f >>> print)
+}
+
+object Printer {
+
+  def apply[A](f: A => String): Printer[A] = new Printer[A] {
+    def print(a: A): String = f(a)
+  }
+
+  def konst(str: String): Printer[Unit] =
```
diesalbla: I think that, with `SAM`, this is just 
```Scala
def konst(str: String): Printer[Unit] = _ => str
```

src/main/scala/Printer.scala
```diff
@@ -0,0 +1,66 @@
...
+  def space = konst(" ")
+
+  def newLine = konst("\n")
+
+  val string: Printer[String] = Printer(identity)
+
+  def optional[A](p: Printer[A]): Printer[Option[A]] =
+    Printer(_.fold("")(p.print))
+
+  def mkList[A](p: Printer[A], sep: String): Printer[List[A]] =
```
diesalbla: Following the naming of monadic parsing libraries (which is just the reverse), I would call this [`sepBy`](http://hackage.haskell.org/package/base-4.12.0.0/docs/Text-ParserCombinators-ReadP.html#v:sepBy).

src/main/scala/Printer.scala
```diff
@@ -0,0 +1,66 @@
...
+object Printer {
+
+  def apply[A](f: A => String): Printer[A] = new Printer[A] {
+    def print(a: A): String = f(a)
+  }
+
+  def konst(str: String): Printer[Unit] =
+    Printer(Function.const(str))
+
+  def space = konst(" ")
```
diesalbla: This `def` can be ` val`. Since it is a public method, I would add the type of it.

src/main/scala/Printer.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  def newLine = konst("\n")
```
diesalbla: Also can be a `val`.

src/main/scala/Printer.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  val string: Printer[String] = Printer(identity)
```
diesalbla: Could we add a `val empty: Printer[Unit] = _ => ""`?

src/main/scala/Printer.scala
```diff
@@ -0,0 +1,66 @@
...
+  def mkList[A](p: Printer[A], sep: String): Printer[List[A]] =
+    Printer {
+      case Nil => ""
+      case xs  => xs.map(p.print).mkString(sep)
+    }
+
+  implicit val divisiblePrinter: Decidable[Printer] = new Decidable[Printer] {
+    def contramap[A, B](fa: Printer[A])(f: B => A): Printer[B] =
+      Printer(f >>> fa.print)
+    def conquer[A]: Printer[A] =
```
diesalbla: This can refer the `empty` printer from the previous comment.

src/main/scala/Printer.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  def newLine = konst("\n")
+
+  val string: Printer[String] = Printer(identity)
+
+  def optional[A](p: Printer[A]): Printer[Option[A]] =
+    Printer(_.fold("")(p.print))
+
+  def mkList[A](p: Printer[A], sep: String): Printer[List[A]] =
+    Printer {
```
diesalbla: Again, by `SAM`, you do not need to add the `Printer` here. It will take the pattern-match as a literal for `PartialFunction`, which in Scala is a `Function` (with some controversy I know).

src/main/scala/Printer.scala
pepegar: Yep, it can!

src/main/scala/Printer.scala
pepegar: I can't use SAM in the library, since I crosscompile it to 2.11 too :(

src/main/scala/Printer.scala
pepegar: I can't introduce SAM inside the lib because crosscompiling to 2.11...

PR #27, opened 2018-11-19
:tada: use @deriveTraverse annotation in our ADTs

pepegar:
This PR:

- uses the new `@deriveTraverse` annotation from Droste. This means that derivation happens mechanically. Also, since we're providing `Traverse` instead of `Functor`, we will be able to use monadic recursion schemes as well!

- introduces smart constructors to avoid scala inferring the type of the case in the ADT instead of the type of the actual ADT.

- Checks laws for created instances

- Adds instance for `Eq` in schemas.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -100,7 +100,8 @@ lazy val commonSettings = Seq(
     "io.higherkindness" %% "droste-macros" % V.droste,
     "org.apache.avro"   % "avro"           % V.avro,
     %%("circe-core", V.circe),
-    %%("specs2-core", V.specs2)       % Test,
+    %%("scalacheck", "1.13.5"),
+    %%("specs2-core"      , V.specs2)       % Test,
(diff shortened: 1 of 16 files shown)


andyscott:
Geez, could you put that first headline in bigger/bolder text? 😄

build.sbt
```diff
@@ -100,9 +101,12 @@ lazy val commonSettings = Seq(
     "io.higherkindness" %% "droste-macros" % V.droste,
     "org.apache.avro"   % "avro"           % V.avro,
     %%("circe-core", V.circe),
-    %%("specs2-core", V.specs2)       % Test,
+    %%("scalacheck", V.scalacheck),
```
diesalbla: Should this be " % Test" ?

pepegar: Indeed, good catch Diego!

Issue #30, opened 2018-11-21
Parse proto files


pepegar:
It would be great to have a way to parse protocol buffers in
Skeuomorph!

This task would involve several smaller tasks to be completed (we can
address each one in a separate PR if we like):

- [ ] compile the proto file into its binary format
- [ ] get a `FileDescriptorProto` out of it
- [ ] convert all message declarations into `ProtobufF` declarations.
- [ ] create a protocol describing a Protobuf file.
- [ ] create transformation between Protobuf's protocol and Mu's
      protocol.

## Compile proto file

To do this we can use [protoc-jar][protoc] as a library.

## Convert proto bytes into a FileDescriptorProto

`FileDescriptorProto` comes from protobuf-java library, the oficial
SDK for protobuf in Java.  The idea would be to use
`FileDescriptorProto.parseFrom` to convert from the bytes generated
with protoc compilation into something with more semantics, like a
`FileDescriptorProto`.

## Convert message declarations to ProtobufF

This is done with a Droste's `Coalgebra`, and the process would be
similar to the one we are doing for avro [here][fromAvro].

## Create a skeuomorph.protobuf.Protocol

This should be similar to [Avro's][AvroProtocol].

## Conversion Protobuf -> Mu

This should be similar to [Mu's fromAvroProtocol][fromAvroProtocol]

[protoc]: https://github.com/os72/protoc-jar#usage---api
[FileDescriptorProto.parseFrom]: https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/DescriptorProtos.FileDescriptorProto#parseFrom-byte:A-
[fromAvro]: https://github.com/higherkindness/skeuomorph/blob/master/src/main/scala/avro/schema.scala#L155
[AvroProtocol]: https://github.com/higherkindness/skeuomorph/blob/master/src/main/scala/avro/Protocol.scala
[fromAvroProtocol]: https://github.com/higherkindness/skeuomorph/blob/master/src/main/scala/mu/protocol.scala#L44

Thanks @AdrianRaFo for https://github.com/AdrianRaFo/proton, from which I pulled the inspiration to create this issue _a la_ skeuomorph

AdrianRaFo:
Proton uses `sbt-protoc` and `scalapb` behind the scenes to read the proto file so it can be helpful to discover how to proceed in some cases

juanpedromoreno:
This has been completed by https://github.com/higherkindness/skeuomorph/pull/66

PR #59, opened 2019-01-09
Protobuf compilation

rlmark:
Addresses parts 1, 2, and 3 of #30 

What this PR does broadly:

Enables Scala code printing from a Protobuf file for Proto 3.

Features of Protobuf that it does not support: 

* Proto 2. 
* Protobuf files that import other Protobuf files 
* Proto messages have messages defined inside of themselves for scoping (like a closure, proto messages that refer to other messages, importantly, are supported).  
* Note: it you are Windows user, the /tmp directory is different for you

Features that it does not support YET: 

* The Mu service code generation. That will have to come in a bit. 
* We should ultimately use the `@deriveTraverse` annotation to create a traverse instance for ProtobufF. Currently there might be some features in ProtobufF that need to be supported by the deriveTraverse functionality in Droste. 

To see it in action: 

Run the Playground app to see Skeuomorph read in a Protobuf file and then turn it into a printed representation of Scala code. 

Areas for future development: 

* All fields in Proto 3 are Optional by default, therefore, perhaps all the Scala code should be expressed in terms of Options. If this is needed, we would probably need to do a transformation like the one written for TRepeated types. 
* Fail fast when we encounter a file for Proto 2. We should immediately tell the user and not do any more computation if their proto file is a proto 2 file.
* Other ideas welcome!


src/main/resources/sampleProto.proto
```diff
@@ -0,0 +1,58 @@
+syntax = "proto3";
```
rlmark: This is just a sample proto file for the playground app.

src/main/scala/higherkindness/skeuomorph/Playground.scala
```diff
@@ -0,0 +1,64 @@
+/*
```
rlmark: This file contains the proof of concept code

src/main/scala/higherkindness/skeuomorph/mu/schema.scala
```diff
@@ -43,8 +44,9 @@ object MuF {
   final case class TOption[A](value: A)                              extends MuF[A]
   final case class TEither[A](left: A, right: A)                     extends MuF[A]
   final case class TList[A](value: A)                                extends MuF[A]
-  final case class TMap[A](value: A)                                 extends MuF[A]
+  final case class TMap[A](keyTpe: Option[A], value: A)              extends MuF[A]
```
rlmark: Original support for Scala Maps had pinned the key to being a String due to Avro limitations, but Protobuf supports variable key types, therefore if you create a TMap with None key type, you get a String printed out as Scala code, whereas Some(type) renders any type. It would be nice to have Avro reflect that its Map key type is a String and avoid this Option in the MuF ADT, but I was unable to do this.

src/main/scala/higherkindness/skeuomorph/mu/schema.scala
```diff
@@ -43,8 +44,9 @@ object MuF {
...
   final case class TGeneric[A](generic: A, params: List[A])          extends MuF[A]
+  final case class TContaining[A](values: List[A])                   extends MuF[A]
```
rlmark: This needed to be added for Scala code-printing because Protobuf's messages and enums come in as a Scala List of values. Perhaps there is a different way of doing this without adding another type to MuF, but I found this useful.

src/main/scala/higherkindness/skeuomorph/protobuf/schema.scala
```diff
@@ -16,111 +16,322 @@
...
+import cats.implicits._
+import com.google.protobuf.descriptor.{EnumOptions, FieldDescriptorProto, UninterpretedOption}
+import com.google.protobuf.descriptor.UninterpretedOption.NamePart
+import higherkindness.skeuomorph.protobuf.ProtobufF.Option
+import qq.droste.Coalgebra
+import qq.droste.util.DefaultTraverse
+import scalapb.descriptors.{ScalaType, _}
 
-import qq.droste.macros.deriveTraverse
+sealed trait FieldF[A] {
```
rlmark: Field needs to be a trait because a Field can either be a normal message field I.e. 
```
string name = 1;
```
but it can also be a `oneOf` block 
```
oneof id {
   string name = 1; 
   int32 number = 2;
}
```

having field be an ADT also opens the door to eventually support nested message definitions for protobuf files.

src/main/scala/higherkindness/skeuomorph/protobuf/schema.scala
```diff
@@ -16,111 +16,322 @@
...
-  final case class TSint64[A]()                extends ProtobufF[A]
-  final case class TFixed32[A]()               extends ProtobufF[A]
-  final case class TFixed64[A]()               extends ProtobufF[A]
-  final case class TSfixed32[A]()              extends ProtobufF[A]
-  final case class TSfixed64[A]()              extends ProtobufF[A]
-  final case class TBool[A]()                  extends ProtobufF[A]
-  final case class TString[A]()                extends ProtobufF[A]
-  final case class TBytes[A]()                 extends ProtobufF[A]
-  final case class TNamedType[A](name: String) extends ProtobufF[A]
-  final case class TRequired[A](value: A)      extends ProtobufF[A]
```
rlmark: required does not appear to be a valid annotation for proto 3

src/main/scala/higherkindness/skeuomorph/protobuf/schema.scala
```diff
@@ -16,111 +16,322 @@
...
-  final case class TOptional[A](value: A)      extends ProtobufF[A]
```
rlmark: The lack of TOptional in ProtobufF is the decision I feel least confident in. All fields in proto 3 are by default, optional, so maybe this should be added back in.

src/test/scala/higherkindness/skeuomorph/instances.scala
```diff
@@ -18,32 +18,293 @@ package higherkindness.skeuomorph
...
 import avro.AvroF
-import protobuf.ProtobufF
+import protobuf.{FieldF, ProtobufF}
+import qq.droste.Basis
+import scalapb.UnknownFieldSet
+import scalapb.descriptors.FileDescriptor
 
 import scala.collection.JavaConverters._
 
 object instances {
```
rlmark: It might be worth it to have separate instances objects for each IDL: one for protobuf, one for avro, one for open api, etc.

src/test/scala/higherkindness/skeuomorph/instances.scala
```diff
@@ -18,32 +18,293 @@ package higherkindness.skeuomorph
...
+        options = None,
+        sourceCodeInfo = None,
+        syntax = Some("proto3")
+      )
+  }
+
+  implicit lazy val baseDescriptorArbitrary: Arbitrary[FileDescriptor] = Arbitrary {
+    for {
+      sampleFileDescriptorProto <- Gen.lzy(sampleFileDescriptorProto.arbitrary)
+    } yield FileDescriptor.buildFrom(sampleFileDescriptorProto, Nil)
```
rlmark: An important note about these generators is that they build up the raw "proto" versions of classes and here at the end transform them into a complete FileDescriptor because the all the constructors in the BaseDescriptor sealed trait are marked package private. This is also why the generators are all tuned with small integer values and are generated lazily: I was trying to minimize a performance hit for such a large generated structure. I'm definitely open to other ideas here!

src/test/scala/higherkindness/skeuomorph/protobuf/ProtobufSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+    case ProtobufF.TNamedType(_)               => true
+  }
+
+  def fieldTest(fieldDesc: BaseDescriptor, target: Either[Type, ProtobufF[Boolean]]): Boolean =
+    fieldDesc match {
+      case f: FileDescriptor =>
+        (f.enums ++ f.messages).forall(
+          d =>
+            // These subsequent hylomorphisms ensure we actually test leaf nodes
+            scheme.hylo(checkProto(d), ProtobufF.fromProtobuf)(implicitly[Functor[ProtobufF]])(d))
```
rlmark: Ok, here's where my testing solution got a bit... creative. I'm hoping someone with more recursion scheme knowledge knows of a way to make sure that all levels of the ProtobufF ADT are tested given a FileDescriptor, as the only way I could get these tests to fail and succeed as expected was with subsequent calls to a hylomorphism.

src/main/scala/higherkindness/skeuomorph/Playground.scala
pepegar: Can we move this proof of concept to a `.md` file in the docs?

src/main/scala/higherkindness/skeuomorph/protobuf/schema.scala
AdrianRaFo: On proto 3 all fields are optional an there is no option to set one as required. I solved that on Proton creating a Null message and doing a `oneOf` with that to mark it as optional(similar to avro handling)

rafaparadela:
Closed in favor of https://github.com/higherkindness/skeuomorph/pull/66

PR #62, opened 2019-01-22
Some Proto compilation enhancements 

rafaparadela:
This PR enriches the great work made by @rlmark at #59 

It basically let the proto parser:

- To compiling proto2 definitions
- To dig into `import`s definitions


rafaparadela:
Closed in favor of #66

PR #81, opened 2019-04-16
Adds the parameters to an operation

fedefernandez:


pepegar:
Hey! This looks like a needed addition, thanks Fede!

I think that, in the case of the parameter ADT, you want the occurrences of Json Schema types to be typed as A, not JsonSchemaF[A].

Sorry, I’m on my phone, but I’ll try to catch-up with this later on

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+  implicit val referenceDecoder: Decoder[Reference] = Decoder.forProduct1(s"$$ref")(Reference.apply)
+
+  implicit def orReferenceDecoder[A: Decoder]: Decoder[Either[A, Reference]] =
+    Decoder[Reference].map(_.asRight[A]) orElse Decoder[A].map(_.asLeft[Reference])
+
+  private def basicJsonSchemaDecoder: Decoder[JsonSchemaF.Fixed] = {
+    import JsonSchemaF.Fixed._
+    Decoder.forProduct1[String, String]("type") { identity }.emap {
+      case "integer"  => integer().asRight
+      case "long"     => long().asRight
```
pepegar: Same here:

```suggestion
      case "long"     => long[A].embed.asRight
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+      case "float"    => float().asRight
```
pepegar:
```suggestion
      case "float"    => float[A].embed.asRight
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+      case "double"   => double().asRight
```
pepegar:
```suggestion
      case "double"   => double[A].embed.asRight
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+
+object JsonDecoders {
+
+  implicit val referenceDecoder: Decoder[Reference] = Decoder.forProduct1(s"$$ref")(Reference.apply)
+
+  implicit def orReferenceDecoder[A: Decoder]: Decoder[Either[A, Reference]] =
+    Decoder[Reference].map(_.asRight[A]) orElse Decoder[A].map(_.asLeft[Reference])
+
+  private def basicJsonSchemaDecoder: Decoder[JsonSchemaF.Fixed] = {
+    import JsonSchemaF.Fixed._
```
pepegar:
```suggestion
    import JsonSchemaF._
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+    Decoder.forProduct1[String, String]("type") { identity }.emap {
+      case "integer"  => integer().asRight
+      case "long"     => long().asRight
+      case "float"    => float().asRight
+      case "double"   => double().asRight
+      case "string"   => string().asRight
```
pepegar:
```suggestion
      case "string"   => string[A].embed.asRight
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+      case "byte"     => byte().asRight
```
pepegar:
```suggestion
      case "byte"     => byte[A].embed.asRight
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+      case "binary"   => binary().asRight
```
pepegar:
```suggestion
      case "binary"   => binary[A].embed.asRight
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+      case "boolean"  => boolean().asRight
```
pepegar:
```suggestion
      case "boolean"  => boolean[A].embed.asRight
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+      case "date"     => date().asRight
```
pepegar:
```suggestion
      case "date"     => date[A].embed.asRight
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+      case "datetime" => dateTime().asRight
```
pepegar:
```suggestion
      case "datetime" => dateTime[A].embed.asRight
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+      case "password" => password().asRight
```
pepegar:
```suggestion
      case "password" => password[A].embed.asRight
```

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,300 @@
...
+    )(
+      (
+          openapi: String,
+          info: Info,
+          servers: Option[List[Server]],
+          paths: Map[String, Path.ItemObject[A]],
+          components: Option[Components[A]],
+          tags: Option[List[Tag]],
+          externalDocs: Option[ExternalDocs]) =>
+        OpenApi.apply(
```
pepegar: Why the `.apply`?

src/main/scala/higherkindness/skeuomorph/openapi/schema.scala
```diff
@@ -15,34 +15,40 @@
...
-
+  final case class Reference(ref: String)
   final case class OpenApi[A](
       openapi: String,
       info: Info,
       servers: List[Server],
       paths: Map[String, Path.ItemObject[A]],
       components: Option[Components[A]],
-      tags: Option[List[Tag]],
+      tags: List[Tag],
```
pepegar: :+1:

build.sbt
```diff
@@ -98,13 +98,15 @@ lazy val commonSettings = Seq(
...
-    "io.higherkindness"   %% "droste-macros" % V.droste,
-    "org.apache.avro"     % "avro"           % V.avro,
-    "com.github.os72"     % "protoc-jar"     % V.protoc,
-    "com.google.protobuf" % "protobuf-java"  % V.protobuf,
+    "io.higherkindness"   %% "droste-core"       % V.droste,
+    "io.higherkindness"   %% "droste-macros"     % V.droste,
+    "org.apache.avro"     % "avro"               % V.avro,
+    "com.github.os72"     % "protoc-jar"         % V.protoc,
+    "com.google.protobuf" % "protobuf-java"      % V.protobuf,
+    "io.circe"            %% "circe-testing"     % V.circe % Test,
```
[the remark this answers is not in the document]
BeniVF: Adding that library could be very useful for testing!

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
BeniVF: Previously it was apply however now it is not required

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,309 @@
...
+object JsonDecoders {
+
+  implicit val referenceDecoder: Decoder[Reference] = Decoder.forProduct1(s"$$ref")(Reference.apply)
+
+  implicit def orReferenceDecoder[A: Decoder]: Decoder[Either[A, Reference]] =
+    Decoder[Reference].map(_.asRight[A]) orElse Decoder[A].map(_.asLeft[Reference])
+
+  private def basicJsonSchemaDecoder[A: Embed[JsonSchemaF, ?]]: Decoder[A] = {
+    import JsonSchemaF._
+    Decoder.forProduct1[String, String]("type") { identity }.emap {
```
diesalbla: Should this correspond to the [Data Types section](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#data-types) of the OpenApi specification?
If so, there are some discrepancies: for instance, there is no `long` type (only the `integer` type with the `int32` or `int64` formats. Also, the `datetime` is actually spelled `date-time`.

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,309 @@
...
+    Decoder.instance(c =>
+      for {
+        values <- c.downField("enum").as[List[String]]
+        _      <- validateType(c, "string")
+      } yield JsonSchemaF.enum[A](values).embed)
+
+  private def objectJsonSchemaDecoder[A: Embed[JsonSchemaF, ?]]: Decoder[A] =
+    Decoder.instance { c =>
+      for {
+        required <- c.downField("required").as[List[String]]
```
diesalbla: Would this _require_, from the input schema given to the parsing, to always have the `required` field, as a list of strings (field names)?

In the OpenAPI Spec, that is not always needed.

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
BeniVF: Very good point, this was my mistake because I followed the previous Encoders implementation.

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,311 @@
...
+object JsonDecoders {
+
+  implicit val referenceDecoder: Decoder[Reference] = Decoder.forProduct1(s"$$ref")(Reference.apply)
+
+  implicit def orReferenceDecoder[A: Decoder]: Decoder[Either[A, Reference]] =
+    Decoder[Reference].map(_.asRight[A]) orElse Decoder[A].map(_.asLeft[Reference])
+
+  private def basicJsonSchemaDecoder[A: Embed[JsonSchemaF, ?]]: Decoder[A] = {
+    import JsonSchemaF._
+    Decoder.forProduct2[(String, Option[String]), String, Option[String]]("type", "format") { Tuple2.apply }.emap {
```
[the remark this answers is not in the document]
BeniVF: I don't think we can because we require  **fail** that's why I am using `emap`.

The product method requires a pure function that given two arguments goes to the **Target**, where **Target** is the final object (not a _option_ or _either_) 

`final def forProduct2[Target, A0, A1](nameA0: String, nameA1: String)(f: (A0, A1) => Target)(implicit decodeA0: Decoder[A0], decodeA1: Decoder[A1]): Decoder[Target]`

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -0,0 +1,311 @@
...
+      } yield JsonSchemaF.enum[A](values).embed)
+
+  private def objectJsonSchemaDecoder[A: Embed[JsonSchemaF, ?]]: Decoder[A] =
+    Decoder.instance { c =>
+      for {
+        required <- c.downField("required").as[Option[List[String]]]
+        properties <- c
+          .downField("properties")
+          .as[Map[String, A]] {
+            implicit lazy val x = jsonSchemaDecoder[A]
```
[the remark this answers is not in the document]
BeniVF: Regarding **docs**, we have a reference in the `schema` package object. However, maybe we could introduce the reference in the relevant decoders as you mentioned.  

Regarding  **tests**, most of the examples I am using in my current tests are based on spec. But it could be nice to introduce these examples. One of the things I was thinking to do is introduce them into the property based tests through `Arbitrary[Json]` and then we could support all the `CodecTests` laws. However,  just keep in mind that we are not supporting the whole open api spec at the moment.

PR #82, opened 2019-04-23
Introduce schema comparison

vil1:
This PR introduces a mechanism for comparing two schemas.

Schemas are compared asymmetrically: one is considered the "writer" and the other the "reader". The comparison builds a result that can be either:
* a `Match` with a (possibly empty) list of *compatible transformations* from `writer` to `reader`
* a `Mismatch` with also a list of compatibility violations  

A compatible transformation is a difference between `writer` and `reader` such that message written using `writer` remain compatible with `reader`.

In a first iteration, I've decided to remain as close as possible to [avro's spec](https://avro.apache.org/docs/1.8.2/spec.html#Schema+Resolution) in the set of accepted transformations. More might be implemented later.


src/main/scala/higherkindness/skeuomorph/mu/comparison/ComparisonResult.scala
```diff
@@ -0,0 +1,71 @@
...
+  final case class Different(relativePath: Path)          extends Incompatibility
+  final case class UnionMemberRemoved(relativePath: Path) extends Incompatibility
+
+  implicit val incompatibilityShow: Show[Incompatibility] = Show.show {
+    case Different(p)          => p.show ++ ": !!DIFFERENT!!"
+    case UnionMemberRemoved(p) => p.show ++ ": union member not found in reader schema"
+  }
+}
+
+sealed trait ComparisonResult[T] {
```
diesalbla: At _first sight_, I can see a more simple representation for the `ComparisonResult` datatype, but just using a plain case class: 
```Scala
final case class ComparisonResult[T](
  transformations: List[Transformation[T]], 
  discrepancies: List[Incompatibility]){

  def isMatch: Boolean = discrepancies.isEmpty
}
```
With this definition, the monoid instance would simply consists of adding both lists, without the match cases. 

However, this is a first sight. Perhaps are there any problems of following that approach?

vil1: As a matter of fact, I pondered such refactoring while writing some scaladoc yesterday. It could in fact be made even simpler
```scala
type ComparisonResult[T] = (List[Transformation[T]], List[Incompatibility])
``` 
and the `Monoid` instance would be for free.

src/main/scala/higherkindness/skeuomorph/mu/comparison/Transformation.scala
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package higherkindness.skeuomorph.mu.comparison
+
+import cats.Show
+import cats.syntax.show._
+
+sealed trait Transformation[T]
+
+object Transformation {
```
pepegar: I would name all cases of this ADT using nouns:

src/main/scala/higherkindness/skeuomorph/mu/comparison/Transformation.scala
```diff
@@ -0,0 +1,56 @@
...
+  final case class NumericWiddening[T](relativePath: Path, from: T, to: T)  extends Transformation[T]
```
pepegar:
```suggestion
  final case class NumericWidening[T](relativePath: Path, from: T, to: T)  extends Transformation[T]
```

src/main/scala/higherkindness/skeuomorph/mu/comparison/Transformation.scala
```diff
@@ -0,0 +1,56 @@
...
+  final case class StringConversion[T](relativePath: Path, from: T, to: T)  extends Transformation[T]
+  final case class Addition[T](relativePath: Path, added: T)                extends Transformation[T]
+  final case class Removal[T](relativePath: Path, removed: T)               extends Transformation[T]
+  final case class MadeOptional[T](relativePath: Path)                      extends Transformation[T]
+  final case class PromotedToEither[T](relativePath: Path, either: T)       extends Transformation[T]
+  final case class PromotedToCoproduct[T](relativePath: Path, coproduct: T) extends Transformation[T]
+  final case class CoproductWiddening[T](relativePath: Path, coproduct: T)  extends Transformation[T]
```
pepegar:
```suggestion
  final case class CoproductWidening[T](relativePath: Path, coproduct: T)  extends Transformation[T]
```

src/main/scala/higherkindness/skeuomorph/mu/comparison/Transformation.scala
```diff
@@ -0,0 +1,56 @@
...
+
+sealed trait Transformation[T]
+
+object Transformation {
+  final case class NumericWiddening[T](relativePath: Path, from: T, to: T)  extends Transformation[T]
+  final case class StringConversion[T](relativePath: Path, from: T, to: T)  extends Transformation[T]
+  final case class Addition[T](relativePath: Path, added: T)                extends Transformation[T]
+  final case class Removal[T](relativePath: Path, removed: T)               extends Transformation[T]
+  final case class MadeOptional[T](relativePath: Path)                      extends Transformation[T]
+  final case class PromotedToEither[T](relativePath: Path, either: T)       extends Transformation[T]
```
pepegar:
```suggestion
  final case class EitherPromotion[T](relativePath: Path, either: T)       extends Transformation[T]
```

src/main/scala/higherkindness/skeuomorph/mu/comparison/Transformation.scala
```diff
@@ -0,0 +1,56 @@
...
+  final case class PromotedToCoproduct[T](relativePath: Path, coproduct: T) extends Transformation[T]
```
pepegar:
```suggestion
  final case class CoproductPromotion[T](relativePath: Path, coproduct: T) extends Transformation[T]
```

src/main/scala/higherkindness/skeuomorph/mu/comparison/Comparison.scala
```diff
@@ -0,0 +1,283 @@
...
+        case (TContaining(a), TContaining(b)) =>
+          CompareList(zipLists(path, a, b, Alternative)) // TODO what's the semantics?
+        case (TEither(l1, r1), TEither(l2, r2)) =>
+          CompareList(List((path / LeftBranch, l1.some, l2.some), (path / RightBranch, r1.some, r2.some)))
+        case (TGeneric(g, p), TGeneric(g2, p2)) =>
+          CompareList((path / GenericType, g.some, g2.some) :: zipLists(path, p, p2, GenericParameter))
+        case (TCoproduct(i), TCoproduct(i2)) =>
+          AlignUnionMembers(
+            i.zipWithIndex
+              .map {
```
diesalbla: In order to trim down the complexity for this case (essentially MxN cases), perhaps we could add a quick "canFit" predicate, one which would approximate the full "is compatible" case. Approximate, in this case, would mean that if `canFit(p._1, p._2)` is false, that implies that `p._1` cannot be compatible with `p._2`. The benefit of course would be that `canFit` would only look at the top level (not recursive).

src/main/scala/higherkindness/skeuomorph/mu/comparison/Transformation.scala
```diff
@@ -0,0 +1,56 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package higherkindness.skeuomorph.mu.comparison
+
+import cats.Show
+import cats.syntax.show._
+
+sealed trait Transformation[T]
```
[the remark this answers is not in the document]
vil1: should be better now

src/main/scala/higherkindness/skeuomorph/mu/comparison/Comparison.scala
vil1: I think must indeed be addressed, but should be in another PR. 

Beside the approach you suggest, we should be able to avoid a whole bunch of unnecessary comparison by making the traversal monadic (`hyloM`) or using another scheme (maybe `apo` but I'm not completely sure about that).

PR #61, opened 2019-01-17
WIP: migrate datatypes to UAST

pepegar:

PR #87, opened 2019-04-24
Atomic

diesalbla:
Refactor our schema ADT, by making a special _Kind_ of atomic schema types. This should help a bit. 

Using Scala enums, but enumeratum could be used instead.


pepegar:
This looks good but there’s an ongoing effort to migrate all ASTs to the same ADT in https://github.com/higherkindness/skeuomorph/pull/61 and adding this will only make that harder. Do you need this now? Or can we integrate it once #61 is ready?

PR #92, opened 2019-05-24
Make fields optional for non-primitive protobuf fields

noelmarkham:
Added a new field, TOptionalNamedType for https://github.com/higherkindness/mu/issues/612 - Protobuf requiring optional fields for non-primitive types.

Would appreciate feedback on this; I feel this can't be everything I need to do for this to be complete.

### DIFF
--- a/src/main/scala/higherkindness/skeuomorph/avro/Protocol.scala
+++ b/src/main/scala/higherkindness/skeuomorph/avro/Protocol.scala
@@ -81,24 +81,25 @@ object Protocol {
   }
 
   def fromFreesFSchema[T](implicit T: Basis[AvroF, T]): Trans[MuF, AvroF, T] = Trans {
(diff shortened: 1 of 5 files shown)


eli-jordan:
I think you will need to have separate handling for proto3 vs proto2 syntax. In proto2 fields are optional by default but can be marked as required. In proto3 all fields are always optional.

pepegar:
@eli-jordan Yep, we're planning to have different printers in the future for proto2 and proto3. Thanks for your comment!

TannerYoung:
Is there a follow up bug explicitly for proto2 vs proto3 support for primitive fields?

The appeal of proto2 is that you can tell whether a field has been explicitly set. For example 
with strings in proto3 it is impossible to differentiate between not set and the empty string, but this is possible with proto2.

It feels a bit odd that this support only extends to non-primitive fields.

PR #96, opened 2019-05-30
Remove fs2 prefix

noelmarkham:
This was causing some headaches downstream in Mu, better to revisit this in its own release I think.

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.0.10-SNAPSHOT"
+version in ThisBuild := "0.0.9.2"
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.0.9.2"
+version in ThisBuild := "0.0.10"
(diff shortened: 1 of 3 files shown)


pepegar:
This looks good.  Also, since this will end up in a macro, I would prefix all `fs2` packages with `_root_`, in order to not make it collide with user's packages called `fs2`.

What do you think?

noelmarkham:
Sounds good - I think that should be done in a separate PR when addressing the `fs2.` prefix

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.0.10-SNAPSHOT"
\ No newline at end of file
+version in ThisBuild := "0.0.9.2"
```
fedefernandez: ☝️ ⚠️

pepegar:
@noelmarkham sounds good

version.sbt
fedefernandez: hey @noelmarkham do you know if sbt-release handles the fourth number well. From [semantic versioning](https://semver.org/) additional labels are separated by `-`, which I think aren't supported by sbt-release.

noelmarkham:
@pepegar I was seeing this as a patch on `0.0.9` - similarly `0.0.9.1` was purely a bump with no real changes as Sonatype had failed.

version.sbt
noelmarkham: `0.0.9.1` was fine, but that was purely a nudge to make the build happen again. Considering this has actual code changes, I'll make it `0.0.10`.

PR #97, opened 2019-06-17
Ff/print open api

BeniVF:


src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
```diff
@@ -0,0 +1,146 @@
...
+  def httpPath[T]: Printer[HttpPath] =
+    (konst("/ ") *< sepBy(string, " / ")).contramap {
+      _.show
+        .split("/")
+        .filter(_.nonEmpty)
+        .map { s =>
+          if (s.startsWith("{") && s.endsWith("}"))
+            s"${s.tail.init}.show"
+          else
+            s""""$s""""
```
noelmarkham: Ultra minor comment: I'd find escaping the quotes easier to read here.

src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala
```diff
@@ -0,0 +1,362 @@
...
+  def operations[T: Basis[JsonSchemaF, ?]]: Printer[(TraitName, Map[String, ItemObject[T]])] =
+    (
+      imports >* newLine,
+      konst("trait ") *< show[TraitName] >* konst("[F[_]] {") >* newLine,
+      space *< space *< konst("import ") *< show[TraitName] >* konst("._") >* newLine,
+      sepBy(method[T], "\n") >* (newLine >* konst("}") >* newLine),
+      clientTypes[T]
+    ).contramapN(operationsTuple[T])
+
+  def un[A, B, C, D](pair: ((A, B), (C, D))): (A, B, C, D) = (pair._1._1, pair._1._2, pair._2._1, pair._2._2)
```
noelmarkham: Feels like Shapeless might be able to do some of these operations out of the box?

src/test/scala/higherkindness/skeuomorph/openapi/helpers.scala
```diff
@@ -0,0 +1,150 @@
...
+  def operation[A](request: Request[A], responses: (String, Response[A])*): Path.Operation[A] =
+    operationFrom(request.asLeft.some, responses.toMap.mapValues(_.asLeft))
+
+  def reference(value: String): Reference = Reference(value)
+
+  private def operationFrom[A](
+      requestBody: Option[Either[Request[A], Reference]],
+      responses: Map[String, Either[Response[A], Reference]]): Path.Operation[A] =
+    Path.Operation[A](
+      List.empty,
```
noelmarkham: This might benefit from named parameters

src/test/scala/higherkindness/skeuomorph/openapi/helpers.scala
```diff
@@ -0,0 +1,150 @@
...
+      allowEmptyValue = allowEmptyValue,
+      schema = schema,
+      required = required)
+
+  def noneMediaType[A] = MediaType[A](None, Map.empty)
+
+  def mediaType[A](a: A) = MediaType[A](a.some, Map.empty)
+
+  def emptyItemObject[A] = Path.ItemObject[A](
+    None,
```
noelmarkham: Same here regarding named parameters

src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala
BeniVF: Initially, I was going to use it, however I didn't because the current project doesn't contain the library in their dependencies.

I am happy to use it if everyone agrees to adding the dependency to this project.

diesalbla: Unless we have a need for something rather big, not puntual, and generalised to "any" number of elements, I would avoid using `shapeless`.

src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
```diff
@@ -0,0 +1,187 @@
...
+      konst("Encoder[") *< show[EncoderName] >* konst("] = "),
+      konst("deriveEncoder[") *< show[EncoderName] >* konst("]"))
+      .contramapN(x => (x, x, x))
+
+  def entityDecoder[T: Basis[JsonSchemaF, ?]]: Printer[DecoderName] =
+    (
+      konst("implicit val ") *< show[DecoderName] >* konst("EntityDecoder: "),
+      konst("EntityDecoder[F, ") *< show[DecoderName] >* konst("] = "),
+      konst("jsonOf[F, ") *< show[DecoderName] >* konst("]"))
+      .contramapN(x => (x, x, x)) //
```
juanpedromoreno:
```suggestion
      .contramapN(x => (x, x, x))
```

src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
```diff
@@ -0,0 +1,187 @@
...
+
+  def entityEncoder[T: Basis[JsonSchemaF, ?]]: Printer[EncoderName] =
+    (
+      konst("implicit val ") *< show[EncoderName] >* konst("EntityEncoder: "),
+      konst("EntityEncoder[F, ") *< show[EncoderName] >* konst("] = "),
+      konst("jsonEncoderOf[F, ") *< show[EncoderName] >* konst("]"))
+      .contramapN(x => (x, x, x)) //
```
juanpedromoreno:
```suggestion
      .contramapN(x => (x, x, x))
```

src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
```diff
@@ -0,0 +1,187 @@
...
+
+  def methodImpl[T: Basis[JsonSchemaF, ?]](withBody: Printer[String]): Printer[OperationWithPath[T]] =
+    (
+      method[T],
+      konst(" = client.expect[") *< responsesTypes >* konst("]"),
+      konst("(Request[F](method = Method.") *< show[HttpVerb],
+      konst(", uri = baseUrl ") *< divBy(httpPath[T], sepBy(queryParameter[T], ""))(unit) >* konst("))"),
+      optional(withBody)).contramapN { x =>
+      val operationId = OperationId(x)
+      (x, operationId -> x._3.responses, x._1, (x._2, x._3.parameters.flatMap(_.left.toOption).collect {
```
juanpedromoreno: `_.left.toOption` wouldn't be the same as `None`?

BeniVF: if you do `_.toOption`  it is going to be `None` if the `Either` contains a `Left`.

If you open ammonite in **skeuomorph**, try the follow:
```scala
@ import cats._, cats.implicits._ 
import cats._, cats.implicits._

@ val x = "foo".asLeft 
x: Either[String, Nothing] = Left("foo")

@ x.left.toOption 
res7: Option[String] = Some("foo")

@ val x = "bar".asRight 
x: Either[Nothing, String] = Right("bar")

@ x.left.toOption 
res10: Option[Nothing] = None
```

src/main/scala/higherkindness/skeuomorph/openapi/print.scala
```diff
@@ -0,0 +1,94 @@
...
+  def schema[T: Basis[JsonSchemaF, ?]](name: Option[String] = None): Printer[T] = {
+    val algebra: Algebra[JsonSchemaF, String] = Algebra { x =>
+      import JsonSchemaF._
+      (x, name) match {
+        case (IntegerF(), _)  => "Int"
+        case (LongF(), _)     => "Long"
+        case (FloatF(), _)    => "Float"
+        case (DoubleF(), _)   => "Double"
+        case (StringF(), _)   => "String"
+        case (ByteF(), _)     => "Array[Byte]"
```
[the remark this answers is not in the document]
BeniVF: The only [reference](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#dataTypes) I can find about `byte` says: 

> base64 encoded characters

and I have interpreted that it could be a sequences of bytes.

The json schema is based on this [specification](https://tools.ietf.org/html/draft-wright-json-schema-00#section-4.2) and for the types they are referring to this [rfc](https://tools.ietf.org/html/rfc7159)

These are the [type test](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/master/tests/draft2019-06/type.json), but I could no find any regarding to `byte`

@pepegar what do you think?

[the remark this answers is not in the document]
pepegar: Yep, you're right, I think it means characters!

BeniVF:
This PR contains the following features:
- Supports the **ammonite**.
- Generates the model based on `#/components/schemas`.
- Generates _trait_ with CRUD operations (_POST_, _GET_, _PUT_, _DELETE)
- Supports a simple **request/response** model, it does not support multiple responses.
- Generates **circe** encoder/decoder.
- Generates _http4s_ implementation with CRUD operations (_POST_, _GET_, _PUT_, _DELETE)
- Supports _http4s_ versions **0.20.x** and **0.18.x**

Issue #85, opened 2019-04-24
The proto and avro printers need to receive a new param for setting the idiomatic gRPC to true


fedefernandez:
Relates to https://github.com/higherkindness/mu/issues/599

We need a way for indicating to the printers that the generated `@service` annotation should be populated with the `namespace` and `methodNameStyle` params.

Ideally, it should be a flag called something like `useIdiomaticRPC` that when true it'll put set the `namespace` equal to `Some($package)` and `methodNameStyle` to `Capitalize`. When false, it can omit those params.

After a talk with @juanpedromoreno, we thought the printer could receive some kind of metadata object with this flag and potentially others.

pepegar:
Sounds good to me. Currently printers are in a very raw state, and they’d need this.

fedefernandez:
This was addressed at https://github.com/higherkindness/skeuomorph/pull/101

PR #101, opened 2019-06-30
CompressionType and useIdiomaticEndpoints for Protobuf and Avro

fedefernandez:
Relates to #85

### DIFF
--- a/src/main/scala/higherkindness/skeuomorph/mu/print.scala
+++ b/src/main/scala/higherkindness/skeuomorph/mu/print.scala
@@ -96,10 +96,10 @@ object print {
    */
   def serviceTuple[T](
       s: Service[T]
-  ): (SerializationType, CompressionType, String, List[Service.Operation[T]]) =
+  ): (SerializationType, CompressionType, IdiomaticEndpoints, String, List[Service.Operation[T]]) =
     s match {
-      case Service(name, serType, compType, ops) =>
-        (serType, compType, name, ops)
+      case Service(name, serType, compType, idiomEndpoints, ops) =>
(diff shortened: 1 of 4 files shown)


fedefernandez:
@franciscodr FYI

src/main/scala/higherkindness/skeuomorph/mu/print.scala
```diff
@@ -136,6 +136,13 @@ object print {
       case Identity => Right(Identity)
     })
 
+  def idiomaticEndpoints: Printer[IdiomaticEndpoints] =
+    Printer {
+      case IdiomaticEndpoints(Some(pkg), true) => ", namespace = Some(\"" + pkg + "\"), methodNameStyle = Capitalize"
```
AlvaroCaste: out of curiosity: is there no "default" Printer for `Some` or `Option`?

fedefernandez: @AlvaroCaste I think I don't understand you, sorry. Could you please rephrase it?

fedefernandez: Oh, you mean there is no case for `Some` + `false`? Fixed!

AlvaroCaste: no, sorry, I got it wrong. I thought you were matching an `Option` and then I wondered if there would already be a `Printer` for it in order not to rewrite `"Some(\"" + pkg + "\")` (this happens when I've just skimmed it 😅 )

fedefernandez:
Ready for review

PR #102, opened 2019-07-01
Ff/print multiple responses

BeniVF:
This PR contains the following features:

- Supports **multiple responses**:
  - Simple `Either` when there is a failure.
  - `Coproduct` of failures when there are multiple failures.
- Supports **anonymous object at the first level** for request and the responses.
- Adds **optional encoders** for the request body.
- Supports **empty responses** in failures.
- Avoids **duplication of types** using the *operation id* in the response.
- **Refactors** to improve readability and simplifies the API.
- Fixes `fetch` method to support `withBody`

### DIFF
--- a/src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
+++ b/src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
@@ -114,32 +114,32 @@ object print {
       case (_, y, _)                                    => y.asLeft
     }
 
-  def statusResponseImpl[T: Basis[JsonSchemaF, ?]]: Printer[(String, (Either[Response[T], Reference], Int))] =
+  def statusResponseImpl[T: Basis[JsonSchemaF, ?]]: Printer[
+    (OperationId, String, (Either[Response[T], Reference], Int))] =
     (
       konst("case response if response.status.code == ") *< string >* konst(" => "),
       konst("response.as[") *< responseOrType[T] >* konst("]"),
       konst(".map(x => ") *< coproductIf(string >* konst("(x)")) >* konst(".asLeft)")).contramapN {
-      case x @ (status, _) =>
-        un(second(x) {
-          case (t, n) =>
-            val (innerTpe, _) = typesAndSchemas[T](status, t)
-            (t, (innerTpe, innerTpe, n))
-        })
+      case (operationId, status, (t, n)) =>
+        val (innerTpe, _) = typesAndSchemas[T](operationId)(status, t)
+        (status, t, (innerTpe, innerTpe, n))
     }
 
-  def defaultResponseImpl[T: Basis[JsonSchemaF, ?]]: Printer[(Either[Response[T], Reference], Int)] =
+  def defaultResponseImpl[T: Basis[JsonSchemaF, ?]]: Printer[(OperationId, (Either[Response[T], Reference], Int))] =
     (
-      konst("case default => default.as[") *< responseOrType[T] >* konst("]"),
-      konst(".map(x => ") *< coproductIf(konst("UnexpectedErrorResponse(default.status.code, x)")) >* konst(".asLeft)"))
-      .contramapN { x =>
-        second(x)(y => ("UnexpectedErrorResponse", (), y))
+      konst("case default => default.as[") *< string >* konst("]"),
+      konst(".map(x => ") *< coproductIf(string >* konst("(default.status.code, x)")) >* konst(".asLeft)"))
+      .contramapN {
+        case (operationId, (x, n)) =>
+          val (tpe, innerTpe, _) = defaultTypesAndSchemas(operationId, x)
+          (innerTpe, (tpe, tpe, n))
       }
 
-  def responseImpl[T: Basis[JsonSchemaF, ?]]: Printer[(String, (Either[Response[T], Reference], Int))] =
-    (successResponseImpl[T] >|< statusResponseImpl[T] >|< defaultResponseImpl[T]).contramap {
-      case (statusCodePattern(code), (x, _)) if successStatusCode(code) => (x.asLeft).asLeft
-      case ("default", x)                                               => x.asRight
-      case x                                                            => x.asRight.asLeft
+  def responseImpl[T: Basis[JsonSchemaF, ?]]: Printer[(OperationId, String, (Either[Response[T], Reference], Int))] =
+    (successResponseImpl[T] >|< defaultResponseImpl[T] >|< statusResponseImpl[T]).contramap {
+      case (_, statusCodePattern(code), (x, _)) if successStatusCode(code) => (x.asLeft).asLeft
+      case (operationId, "default", x)                                     => (operationId -> x).asRight.asLeft
+      case x                                                               => x.asRight
     }
 
   private def queryParametersFrom[T]: Path.Operation[T] => List[Parameter.Query[T]] =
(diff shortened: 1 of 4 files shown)


src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
```diff
@@ -61,51 +61,147 @@ object print {
...
+  def successResponseImpl[T: Basis[JsonSchemaF, ?]]: Printer[Either[Response[T], Reference]] =
+    (konst("case Successful(response) => response.as[") *< responseOrType[T] >* konst("].map(_.asRight)"))
+      .contramap(identity)
+
+  def coproduct[A](printer: Printer[A]): Printer[(String, A)] =
+    (konst("Coproduct[") *< string >* konst("]"), konst("(") *< printer >* konst(")")).contramapN(identity)
+
+  def coproductIf[A](printer: Printer[A]): Printer[(String, A, Int)] =
+    (printer >|< coproduct(printer)).contramap {
+      case (x, y, numberOfErrors) if numberOfErrors > 1 => (x -> y).asRight
```
noelmarkham: Should this be `>=`?

BeniVF:
This PR contains the following features:

- Supports **multiple responses**:
  - Simple `Either` when there is a failure.
  - `Coproduct` of failures when there are multiple failures.
- Supports **anonymous object at the first level** for request and the responses.
- Adds **optional encoders** for the request body.
- Supports **empty responses** in failures.
- Avoids **duplication of types** using the *operation id* in the response.
- **Refactors** to improve readability and simplifies the API.
- Fixes `fetch` method to support `withBody`

src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
BeniVF: `Coproduct` is only used when there is **more than one failure** and the reason for this is:

- when **no failures** have been defined, the _API_ looks like this:
```scala 
def xxx(): F[Something]
```
e.g. https://github.com/higherkindness/skeuomorph/blob/eb2df599c6c03a71f4e812518972c17c3a0284e7/src/test/scala/higherkindness/skeuomorph/openapi/OpenApiPrintSpecification.scala#L77
- when **a failure** has been defined, the _API_  looks like this:
```scala
def xxx(): F[Either[XXXError, Something]]
```
where `XXXError` is:
```scala
type XXXError = XXXSomeErrorResponse
```
e.g. https://github.com/higherkindness/skeuomorph/blob/eb2df599c6c03a71f4e812518972c17c3a0284e7/src/test/scala/higherkindness/skeuomorph/openapi/OpenApiPrintSpecification.scala#L139
- when **multiple failures** has been defined, the _API_  looks like this:
```scala
def xxx(): F[Either[XXXError, Something]]
```
where `XXXError` is:
```scala
type XXXError = XXXSomeErrorResponse :+: XXXOtherErrorResponse :+: ... :+: CNil
``` 
e.g. https://github.com/higherkindness/skeuomorph/blob/eb2df599c6c03a71f4e812518972c17c3a0284e7/src/test/scala/higherkindness/skeuomorph/openapi/OpenApiPrintSpecification.scala#L220

src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala
```diff
@@ -32,13 +32,18 @@ object print {
   import higherkindness.skeuomorph.openapi.schema.Path._
 
   private val jsonMediaType = "application/json"
+  val statusCodePattern     = """(\d+)""".r
+  def successStatusCode(code: String): Boolean = code match {
```
pepegar: Here you're handling redirections as successes, is that on purpose?

BeniVF: yes, at the moment, it is a way to simplify it. In future we should change this to handle specifically redirections.

src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
```diff
@@ -49,63 +50,160 @@ object print {
...
+      konst("implicit val ") *< string >* konst("EntityEncoder: "),
+      konst("EntityEncoder[F, ") *< tpe[T] >* konst("] = "),
+      konst("jsonEncoderOf[F, ") *< tpe[T] >* konst("]"))
+      .contramapN(x => varTuple(x.value))
+
+  def successResponseImpl[T: Basis[JsonSchemaF, ?]]: Printer[Either[Response[T], Reference]] =
+    (konst("case Successful(response) => response.as[") *< responseOrType[T] >* konst("].map(_.asRight)"))
+      .contramap(identity)
+
+  def coproduct[A](printer: Printer[A]): Printer[(String, A)] =
```
diesalbla: I just wonder if, in the smallest cases, they would not be more approachable with a direct style. For instance, unless I am mistaken, this case can be written as follows: 
```Scala
def coproduct[A](printer: Printer[A]): Printer[(String, A)] = { 
  case (str, a) =>  s"Coproduct[$str](${printer(a)})"
}

File: src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala [hunk 4]
```diff
@@ -244,71 +249,119 @@ object print {
...
+      },
+      _ => (newType, tpe, List(statusCaseClass(tpe)))
+    )
+  }
+
+  def statusTypesAndSchemas[T: Basis[JsonSchemaF, ?]](
+      operationId: OperationId,
+      responseOr: Either[Response[T], Reference]): (String, Option[String], List[String]) = {
+    val tpe = responseOrType.print(responseOr)
+    responseOr.left.toOption
```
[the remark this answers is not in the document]
BeniVF: maybe I will use `fold` in that case, not sure why I did in that way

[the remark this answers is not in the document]
diesalbla: As a pure matter of opinion,  that higher-order functions on `Option` or `Either` should not be used unless they cover one line. Once each case becomes a block of several lines, the context given by a `match`-`case` may be better, and avoids a few braces or parentheses.

src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
BeniVF: For  `printer` you need to call `print` like `printer.print(a)`. However from my opinion and given the context, I find it more consistent, safer and more semantic to use the `printer` approach, even in this simple case.

src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/print.scala
```diff
@@ -49,63 +50,160 @@ object print {
...
+
+  def responseImpl[T: Basis[JsonSchemaF, ?]]: Printer[(OperationId, String, (Either[Response[T], Reference], Int))] =
+    (successResponseImpl[T] >|< defaultResponseImpl[T] >|< statusResponseImpl[T]).contramap {
+      case (_, statusCodePattern(code), (x, _)) if successStatusCode(code) => (x.asLeft).asLeft
+      case (operationId, "default", x)                                     => (operationId -> x).asRight.asLeft
+      case x                                                               => x.asRight
+    }
+
+  private def queryParametersFrom[T]: Path.Operation[T] => List[Parameter.Query[T]] =
+    _.parameters.flatMap(_.left.toOption).collect {
```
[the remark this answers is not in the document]
BeniVF: it's not the same, we need to pattern match on the type as well, don't we?:

```scala
  private def queryParametersFrom[T]: Path.Operation[T] => List[Parameter.Query[T]] =
    _.parameters.collect {
      case Left(x: Parameter.Query[T]) => x
    }
```

noelmarkham:
LGTM!

PR #103, opened 2019-07-02
UAST: CompressionType and useIdiomaticEndpoints for Protobuf and Avro

fedefernandez:
Analogous to #101 but against `uast` branch.


pepegar:
I LOVE YOU FEDE, thanks

PR #115, opened 2019-07-15
Use import root to compile Protobuf files

bilki:
This PR adds a new argument to the `protoc` argument list, that allows finding nested packages when multiple protobuf file imports come from different folders/modules.

It's a breaking change for skeuomorph users like `mu-idlgen-sbt` (if `importRoot` is mandatory, maybe it's a good idea to keep if `Option[String]` and provide a default `None` parameter?), that now have to instance a `ProtoSource` including a common root folder for all proto imports.


fedefernandez:
@bilki is this still mergeable?

bilki:
It is, **but** there is another change in _mu_ that needs to be merged too and then bump their dependency version on skeuomorph to the tag where we are going to release this change. Also this needs to be rebased against master.

PR #119, opened 2019-07-23
Removes the optionality in Coproducts

### LINKED ISSUE #113: Protobuf oneOf to generate coproducts without Option-ality
In `proto3` there is no `required` constraint available.

As a consequence this

```
oneof value {
  A : a = 1;
  B : B = 2;
}
```
generates
```
Option[A] :+: Option[B] :+: CNil
```

Proposing the `Option`-ality is removed to generate `shapeless` `A :+: B :+: CNil` instead.

Rationale:

1. The `Option`-ality seems redundant in the coproduct for evident reasons: if A is present in the `coproduct` then it is not optional.
2. Generates more complex than necessary service apis.
3. Slows down the runtime marshalling.

Given that `skeuomorph` is schema transformation library the user should have more flexibility at the time use from `mu` (time of service api design) and not be bound by the semantics of source IDL (context https://github.com/higherkindness/skeuomorph/issues/91).

Resources
1.  https://stackoverflow.com/questions/42955621/require-a-oneof-in-protobuf
2. https://news.ycombinator.com/item?id=18188519

rafaparadela:
This PR fixes #113 

It converts the `TOption` values within `TCoproduct` into `TRequired`.

### DIFF
--- a/src/main/scala/higherkindness/skeuomorph/mu/Optimize.scala
+++ b/src/main/scala/higherkindness/skeuomorph/mu/Optimize.scala
@@ -52,14 +52,26 @@ object Optimize {
     case other => other
   }
 
+  def nestedOptionInCoproductsTrans[T](implicit T: Basis[MuF, T]): Trans[MuF, MuF, T] = Trans {
+    case TCoproduct(nel) => TCoproduct(nel.map(toRequiredTypes))
+    case other           => other
+  }
+
+  def toRequiredTypesTrans[T]: Trans[MuF, MuF, T] = Trans {
+    case TOption(value) => TRequired(value)
+    case other          => other
+  }
+
   def namedTypesTrans[T]: Trans[MuF, MuF, T] = Trans {
     case TProduct(name, _) => TNamedType[T](name)
     case TSum(name, _)     => TNamedType[T](name)
     case other             => other
   }
 
-  def namedTypes[T: Basis[MuF, ?]]: T => T       = scheme.cata(namedTypesTrans.algebra)
-  def nestedNamedTypes[T: Basis[MuF, ?]]: T => T = scheme.cata(nestedNamedTypesTrans.algebra)
+  def namedTypes[T: Basis[MuF, ?]]: T => T              = scheme.cata(namedTypesTrans.algebra)
+  def nestedNamedTypes[T: Basis[MuF, ?]]: T => T        = scheme.cata(nestedNamedTypesTrans.algebra)
+  def toRequiredTypes[T: Basis[MuF, ?]]: T => T         = scheme.cata(toRequiredTypesTrans.algebra)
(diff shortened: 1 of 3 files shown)


rafaparadela:
Not really. We are just removing the optionality in coproducts, but we might want to remain the options in the rest of the cases.

SemanticBeeng:
> Not really. We are just removing the optionality in coproducts, but we might want to remain the options in the rest of the cases.

Thanks for the PR.
Agreed to above point and was wondering if there are ways to control this from options somehow - thought I saw code suggesting the possibility. :thinking: 

Always related to https://github.com/julien-lafont/protoless and its motivation - _"protobuf is not your domain model"_ - and always feel the need to control the mapping to the more rich domain model.

rafaparadela:
I totally agree with your statement "_always feel the need to control the mapping to the more rich domain model_", but I believe that the original purpose of this library was providing the mechanisms to transform protocols into a canonical IDL without losing information and tending to preserve the nature of the original schema. 

It's great to have this debate, and thanks for bringing it up. 

I think that Skeuomorph should keep being accurate when interpreting the schemas (whatever IDL). If there are not required fields in Proto 3, that is to say, all fields are "optional", then the transformation from Proto 3 to Scala must be sensitive to this circumstance, by using the scala `Option`.

On the other hand, there are strategies that users can follow to express the nullability of a field, to differentiate when a field is missing to preserve backward compatibility or when explicitly a field must be null to express lack to value.

**Proto2**

```proto
message Foo {
    optional int32 bar = 1;
}
```

Here, behind the field `bar` there is a semantic which denotes nullability.

**Proto3**
```proto
message Foo {
   int32 bar = 1;
}
```

Here, `bar` is optional perse, but not nullable. I mean, if the value is not provided, does not fail, but we don't really know if this lack comes from legacy clients that do not know the existence of the field `bar`, or explicitly they want to emit a null value.

But we can express this nullability semantic by:

```proto
message Foo {
  oneof bar {
    bool bar_null = 1; 
    int32 bar_value = 2;
  }
}
```

In my last PR, we are omitting the redundant optionality within coproducts because just adds complexity, nothing else.

Actually, [we used to have some optimization](https://github.com/higherkindness/skeuomorph/blob/master/src/main/scala/higherkindness/skeuomorph/mu/Optimize.scala#L81-L89) in these cases, where Skeuomorph was able to detect when the user is attempting to express this nullability, and the coproduct was transformed into `Option` or `Either`.

In summary, I'm inclined to keep interpreting the Protobuf fields as properties within the scala-option.

SemanticBeeng:
Thank you. Agreed with this thought process as well.

> _"strategies that users can follow to express"_

Quite interested to learn about how to express those as user of`mu`  (or of `skeuomorph` from `mu`) as well as in `skeuomorph` itself.

PR #120, opened 2019-07-26
[WIP] transforming Avro to Protobuf and viceversa

AlvaroCaste:
This is a first idea of what it might be this transformation.

There are, for instance `AvroF.Fixed`, which are not clear how they should be transformed.

Anyway, please take a look in order to discuss a better way to make these transformation :)

### DIFF
--- a/src/main/scala/higherkindness/skeuomorph/avro/Transform.scala
+++ b/src/main/scala/higherkindness/skeuomorph/avro/Transform.scala
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2018-2019 47 Degrees, LLC. <http://www.47deg.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package higherkindness.skeuomorph.avro
+
+import higherkindness.skeuomorph.avro.AvroF.{
+  Field,
+  TArray,
+  TBoolean,
+  TBytes,
+  TDouble,
+  TEnum,
+  TFloat,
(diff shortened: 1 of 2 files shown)


src/main/scala/higherkindness/skeuomorph/avro/Transform.scala
```diff
@@ -0,0 +1,70 @@
...
+    case ProtobufF.TBytes()         => TBytes()
+    case ProtobufF.TNamedType(name) => TNamedType(name)
+    case ProtobufF.TRepeated(value) => TArray(value)
+    case ProtobufF.TEnum(name, symbols, _, aliases) =>
+      TEnum(name, None, aliases.map { case (alias, _) => alias }, None, symbols.map { case (symbol, _) => symbol })
+    case ProtobufF.TMessage(name, fields, _) =>
+      TRecord(name, None, List.empty, None, fields.map(f => Field(f.name, List.empty, None, None, f.tpe)))
+    case ProtobufF.TFileDescriptor(values, name, pkg) =>
+      TRecord(name, Some(pkg), List.empty, None, values.map(tpe => Field("", List.empty, None, None, tpe)))
+    case ProtobufF.TOneOf(_, fields) => TUnion(fields.map(_.tpe))
```
bilki: Not sure about this one, as `oneOf` has a field name (it's a field in fact), and we are translating it directly to an Avro union that obviously discards field name. Maybe `Field` would be a more appropriate translation?

src/main/scala/higherkindness/skeuomorph/avro/Transform.scala
```diff
@@ -0,0 +1,70 @@
...
+    case ProtobufF.TMap(_, values)   => TMap(values)
```
bilki: Same scenario as for `oneOf`.

src/main/scala/higherkindness/skeuomorph/avro/Transform.scala
AlvaroCaste: in that case, it would be a `AvroF.TRecord`, where there is a `List[Field[A]]` since `oneOf` also contains a `NonEmptyList[Field[A]]`. WDYT?

src/main/scala/higherkindness/skeuomorph/avro/Transform.scala
AlvaroCaste: I think `TMap` is different since Protobuf allows us to define key type and Avro just defines key types as String.

PR #90, opened 2019-05-22
Ff/open api to mu

BeniVF:
We are going to pospone the translation from **Open Api to Mu**, until we clarify the requirements.

In this pull request, I added the following features to the **Open Api Spec**:
- _Schemas_ inside _Components_. 
- _Yaml_ support.

PR #121, opened 2019-07-30
Normalize and more support for open api spec

BeniVF:
# Objective
You can find, the objectives for all these PRs here: https://github.com/47deg/marlow/issues/188
> Generating http4s client code based on a Open API specification using http4s v0.20 and 0.18 using circe to encode/decode.

# Normalize
**Open api spec** is based on _http_ and is not in a specific programming language. Names of different type of elements (e.g.: schema names, property names, reference names, etc.) can contain symbols, numbers, etc. or grammar rules that are not allowed in the target programming language (in this case _scala_) when we are generating code.  This could be problematic because we are generating code that it is not able to compile.
```yaml
components:
  schemas:
    general-error:
      type: object
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string
    category:
      type: object
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
    tag:
      type: object
      properties:
        id:
          type: integer
          format: int64
        tag_name:
          type: string
```

This is going to be another problem for **encoder/decoders** because there is a mismatch between the _http protocol_ and the _scala_ model. This gap needs to be addressed during the generation of the **encoders/decoders** code.

# Request headers
 **Open api spec** provides a way to define headers in the request.
```yaml
name: token
in: header
description: token to be passed as a header
required: true
schema:
  type: array
  items:
    type: integer
    format: int64
style: simple
```
# Parameters
 **Open api spec** provides a way to define parameter at different levels.
## Components
We can share parameters for different requests in your definition and we can refer to them using `#/components/parameters/{param_name}` when _param_name_ is the name of the parameter.
```yaml
components:
 parameters:
    offsetParam:      # Can be referenced via '#/components/parameters/offsetParam'
      name: offset
      in: query
      description: Number of items to skip before returning the results.
      required: false
      schema:
        type: integer
        format: int32
        minimum: 0
        default: 0
    limitParam:       # Can be referenced as '#/components/parameters/limitParam'
      name: limit
      in: query
      description: Maximum number of items to return.
      required: false
      schema:
        type: integer
        format: int32
        minimum: 1
        maximum: 100
        default: 20
```

## Path Item Object
We can share parameters for all the operations defined at the path item object level.

```yaml
get:
  description: Returns pets based on ID
  summary: Find pets by ID
  operationId: getPetsById
  responses:
    '200':
      description: pet response
      content:
        '*/*' :
          schema:
            type: array
            items:
              $ref: '#/components/schemas/Pet'
    default:
      description: error payload
      content:
        'text/html':
          schema:
            $ref: '#/components/schemas/ErrorModel'
parameters:
- name: id
  in: path
  description: ID of pet to use
  required: true
  schema:
    type: array
    style: simple
    items:
      type: string 
```
# Implicit Time query parameter
When we have a query parameter that is from the`java.time` package it is required to provide implicitly `QueryParamEncoder`. For the following example, we should provide `QueryParamEncoder[java.time.LocalDate]` :
```scala
def getUsers(
            startDate: Option[java.time.LocalDate]): F[List[User]] =
          client.expect[List[User]](
            Request[F](
              method = Method.GET,
              uri = baseUrl / "v0" / "users" +?? ("startDate", startDate)
            ))
```

# Sum types
**Open api spec** provides a way to define a sum type. Here is an example.
```yaml
 - Vehicle:
      oneOf:
        - $ref: "#/components/schemas/Car"
        - $ref: "#/components/schemas/Boat"
        - $ref: "#/components/schemas/Ship"
```

This needs to be supported at all levels.

# Previous PRs
- https://github.com/higherkindness/skeuomorph/pull/116
- https://github.com/higherkindness/skeuomorph/pull/109
- https://github.com/higherkindness/skeuomorph/pull/102
- https://github.com/higherkindness/skeuomorph/pull/97
- https://github.com/higherkindness/skeuomorph/pull/90
- https://github.com/higherkindness/skeuomorph/pull/81

### DIFF
--- a/src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
+++ b/src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
@@ -63,6 +63,12 @@ object JsonDecoders {
         _      <- validateType(c, "string")
       } yield JsonSchemaF.enum[A](values).embed)
 
+  private def sumJsonSchemaDecoder[A: Embed[JsonSchemaF, ?]]: Decoder[A] =
+    Decoder.instance(c =>
+      for {
+        values <- c.downField("oneOf").as[List[A]]
+      } yield JsonSchemaF.sum[A](values).embed)
+
   private def objectJsonSchemaDecoder[A: Embed[JsonSchemaF, ?]]: Decoder[A] =
     Decoder.instance { c =>
       def propertyExists(name: String): Decoder.Result[Unit] =
@@ -73,7 +79,7 @@ object JsonDecoders {
       def isObject: Decoder.Result[Unit] =
         validateType(c, "object") orElse
           propertyExists("properties") orElse
-          propertyExists("allOf") orElse propertyExists("oneOf")
+          propertyExists("allOf")
       for {
         _        <- isObject
         required <- c.downField("required").as[Option[List[String]]]
@@ -99,6 +105,7 @@ object JsonDecoders {
 
   implicit def jsonSchemaDecoder[A: Embed[JsonSchemaF, ?]]: Decoder[A] =
     referenceJsonSchemaDecoder orElse
+      sumJsonSchemaDecoder orElse
       objectJsonSchemaDecoder orElse
       arrayJsonSchemaDecoder orElse
       enumJsonSchemaDecoder orElse
--- a/src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
+++ b/src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
@@ -64,10 +64,7 @@ object JsonDecoders {
       } yield JsonSchemaF.enum[A](values).embed)
 
   private def sumJsonSchemaDecoder[A: Embed[JsonSchemaF, ?]]: Decoder[A] =
-    Decoder.instance(c =>
-      for {
-        values <- c.downField("oneOf").as[List[A]]
-      } yield JsonSchemaF.sum[A](values).embed)
+    Decoder.instance(_.downField("oneOf").as[List[A]].map(JsonSchemaF.sum[A](_).embed))
 
   private def objectJsonSchemaDecoder[A: Embed[JsonSchemaF, ?]]: Decoder[A] =
     Decoder.instance { c =>
--- a/src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
+++ b/src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
@@ -280,16 +280,22 @@ object JsonDecoders {
               servers.getOrElse(List.empty))))
 
   implicit def componentsDecoder[A: Decoder]: Decoder[Components[A]] =
-    Decoder.forProduct3(
+    Decoder.forProduct4(
       "schemas",
       "responses",
-      "requestBodies"
+      "requestBodies",
+      "parameters"
     )(
       (
           schemas: Option[Map[String, A]],
           responses: Option[Map[String, Either[Response[A], Reference]]],
-          requestBodies: Option[Map[String, Either[Request[A], Reference]]]) =>
-        Components(schemas.getOrElse(Map.empty), responses.getOrElse(Map.empty), requestBodies.getOrElse(Map.empty)))
+          requestBodies: Option[Map[String, Either[Request[A], Reference]]],
+          parameters: Option[Map[String, Either[Parameter[A], Reference]]]) =>
+        Components(
+          schemas.getOrElse(Map.empty),
+          responses.getOrElse(Map.empty),
+          requestBodies.getOrElse(Map.empty),
+          parameters.getOrElse(Map.empty)))
 
   implicit def openApiDecoder[A: Decoder]: Decoder[OpenApi[A]] =
     Decoder.forProduct7(
--- a/src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
+++ b/src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
@@ -238,7 +238,7 @@ object JsonDecoders {
         ))
 
   implicit def itemObjectDecoder[A: Decoder]: Decoder[Path.ItemObject[A]] =
-    Decoder.forProduct12(
+    Decoder.forProduct13(
       "ref",
       "summary",
       "description",
@@ -250,7 +250,8 @@ object JsonDecoders {
       "head",
       "patch",
       "trace",
-      "servers")(
+      "servers",
+      "parameters")(
       (
           (
               ref: Option[String], // $ref
@@ -264,7 +265,9 @@ object JsonDecoders {
               head: Option[Path.Operation[A]],
               patch: Option[Path.Operation[A]],
               trace: Option[Path.Operation[A]],
-              servers: Option[List[Server]]) =>
+              servers: Option[List[Server]],
+              parameters: Option[List[Either[Parameter[A], Reference]]]
+          ) =>
             Path.ItemObject(
               ref,
               summary,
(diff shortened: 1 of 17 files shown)


src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala
```diff
@@ -87,68 +163,61 @@ object print {
...
-    case componentsRegex(name) => Var.tpe(Tpe.apply(name)).some
-    case _                     => none
-  }
+  private def requestTuple[T: Basis[JsonSchemaF, ?]](
+      operationId: Http.OperationId,
+      request: Request[T]): Option[VarWithType[T]] =
+    jsonFrom(request.content)
+      .flatMap(
+        x =>
+          x.schema.map(x =>
```
juanpedromoreno: variable shadowing here?

src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala
```diff
@@ -20,62 +20,138 @@ import higherkindness.skeuomorph.Printer
...
+        case Post => "create"
+        case Put  => "update"
+        case x    => x.show.toLowerCase()
+      }
 
-    def methodFrom: HttpVerb => String = {
-      case Post => "create"
-      case Put  => "update"
-      case x    => x.show.toLowerCase()
+      implicit val httpVerbShow: Show[Verb] = Show.show {
```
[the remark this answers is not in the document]
BeniVF: I know the library and it's great, but there are two things:
- should we add a new dependency to this project?
- In this case _enumeratum_ it's not going to work for all the cases. For example, we transform `Post` to `create`, however if we use emumeratum it would transform to `post`, that from the perspective of the end user it's a bit unfriendly.

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
```diff
@@ -277,19 +284,26 @@ object JsonDecoders {
...
-      "requestBodies"
+      "requestBodies",
+      "parameters"
     )(
       (
           schemas: Option[Map[String, A]],
           responses: Option[Map[String, Either[Response[A], Reference]]],
-          requestBodies: Option[Map[String, Either[Request[A], Reference]]]) =>
-        Components(schemas.getOrElse(Map.empty), responses.getOrElse(Map.empty), requestBodies.getOrElse(Map.empty)))
+          requestBodies: Option[Map[String, Either[Request[A], Reference]]],
```
diesalbla: I have noticed that there is little difference, for our purposes, between having a `None` and `Some(Map.empty)`, so perhaps we can remove the `Option` layer. For printing back to `OpenApi`, we could just assume that an empty map is the same as there being no section. 

One inconvenience of this is that this would somehow break the property that encoding/decoding a schema in OpenApi always gives back the "same" schema, but it is debatable if a schema with an empty `requestBodies` section is also the same.

src/main/scala/higherkindness/skeuomorph/openapi/client/http4s/circe/package.scala
```diff
@@ -49,28 +49,51 @@ package object circe {
...
-      space *< space *< (optional(circeDecoder[T]) >|< enumCirceDecoder) >* newLine,
+      space *< space *< (optional(circeEncoder[T]) >|< forProductCirceEncoder[T] >|< enumCirceEncoder >|< sumCirceEncoder) >* newLine,
+      space *< space *< (optional(circeDecoder[T]) >|< forProductCirceDecoder[T] >|< enumCirceDecoder >|< sumCirceDecoder) >* newLine,
       space *< space *< entityEncoder[T] >* newLine,
       space *< space *< entityEncoder[T] >* newLine,
       space *< space *< entityDecoder[T] >* newLine)
       .contramapN {
-        case CaseClassCodecs(name) =>
+        case CaseClassCodecs(name, fields)
+            if fields.forall(field =>
```
diesalbla: I have noticed two things about this case: 
- Perhaps this predicate can be extracted as a function `def isIdentifier(s: String): Boolean`. 
- If you trust [regular expressions](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html), this could be written `field.matches("""^\p{Lower}\p{Alnum}*$""")`, but I admit it may be an overkill.

src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala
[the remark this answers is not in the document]
pepegar: I'd rather not add a new library here

src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala
```diff
@@ -20,62 +20,138 @@ import higherkindness.skeuomorph.Printer
...
-    implicit val httpPathShow: Show[HttpPath] = Show.show(_.value)
+    final case class OperationId private (value: String) extends AnyVal
+
+    object OperationId {
+
+      def apply[T](verb: Verb, path: Path, operation: openapi.schema.Path.Operation[T]): OperationId = {
+        val (paths, varPaths, queries) = Path.splitParts(path)
+        val params                     = varPaths ++ queries
+        OperationId(decapitalize(normalize(operation.operationId.getOrElse {
+          s"${Http.Verb.methodFrom(verb)}${paths.mkString}${if (params.nonEmpty) s"By${params.mkString}"
```
pepegar: I find this line hard to read because of the use of non-escaped double quotes inside the string. Can you use `"""` as a wrapper for the overall string? that would make it easier to understand what's going on for me.

src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala
```diff
@@ -210,8 +284,10 @@ object print {
   }
 
   def defaultTypesAndSchemas[T: Basis[JsonSchemaF, ?]](
-      operationId: OperationId,
+      operationId: Http.OperationId,
       responseOr: Either[Response[T], Reference])(implicit codecs: Printer[Codecs]): (String, String, List[String]) = {
+    def unexpectedErrorName(operationId: Http.OperationId): String =
+      s"${operationId.show}UnexpectedErrorResponse".capitalize
```
pepegar: You can use the `show""` interpolator.  It would fail at compile time when trying to interpolate a value of a type for which there isn't a `Show` instance.  Also, you save the call to `.show`.

```suggestion
          show"$operationId${normalize(name).capitalize}Error".capitalize
```

src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala
```diff
@@ -228,14 +304,16 @@ object print {
...
   def statusTypesAndSchemas[T: Basis[JsonSchemaF, ?]](
-      operationId: OperationId,
+      operationId: Http.OperationId,
       responseOr: Either[Response[T], Reference])(
       implicit codecs: Printer[Codecs]): (String, Option[String], List[String]) = {
     val tpe = responseOrType.print(responseOr)
     responseOr.left.toOption
       .map { response =>
+        def defaultResponseErrorName(operationId: Http.OperationId, name: String): String =
+          s"${operationId.show}${normalize(name).capitalize}Error".capitalize
```
pepegar: Same here, you can use `show""`

src/main/scala/higherkindness/skeuomorph/openapi/client/print.scala
BeniVF: maybe I can use a tmp variable to make it clearest

src/main/scala/higherkindness/skeuomorph/openapi/JsonDecoders.scala
BeniVF: I need to think on this, but maybe if we decide to do this change I will do it in a different PR.

PR #166, opened 2019-11-24
Exhaustive protobuf tests

AdrianRaFo:
This PR shows how protobuf serialization still has failures to be resolved:

**Everything on Proto3 should be optional and we are just generation as optional when using generated messages.**

As Proto2 is also converted to Proto3 we should take some design decisions about this. Also, I cannot find any way to generate an intentional optional with the current design.

**Nested messages classes are not generated. The `BookSeries` model inside the `Series` one is not generated.**

Due to this error, I realized that we are not generating and compiling the scala from the proto we are using. This would help us to detect more bugs in the generation

**Scala keywords are not quoted so scala generated code will not compile**

If we use a scala keyword as the name of a field the generator should quote it using **`** so the generated code can compile

### DIFF
--- a/src/test/scala/higherkindness/skeuomorph/protobuf/ProtobufProtocolSpec.scala
+++ b/src/test/scala/higherkindness/skeuomorph/protobuf/ProtobufProtocolSpec.scala
@@ -32,7 +32,7 @@ class ProtobufProtocolSpec extends Specification with ScalaCheck {
   val root                                      = "/src/test/scala/higherkindness/skeuomorph/protobuf"
   val path                                      = currentDirectory + s"$root/service"
   val importRoot                                = Some(currentDirectory + root)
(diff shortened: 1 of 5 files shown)


fedefernandez:
Would it make sense to add `ignored` to these tests and get that merged? @AdrianRaFo @juanpedromoreno @pepegar thoughts?

AdrianRaFo:
I would not do that because we don't want to lose those tests on other PRs that can cause more issues

juanpedromoreno:
Makes sense to me, @fedefernandez

fedefernandez:
@AdrianRaFo the idea behind merging this is to avoid losing the code and to avoid maintaining the code. Once merged with the ignored keyword they will be seen as a warning, telling us that this is something to be fixed. Does it make sense?

AdrianRaFo:
I think it's too late to do that so I'll close this

PR #178, opened 2019-12-05
Avoid generating invalid Scala identifiers when printing schemas and protocols

### LINKED ISSUE #114: Add a new combinator to escape Scala reserved words
Given a valid schema in any format (mu, protobuf, avro, openapi...), when printing the Scala code, reserved words are not being correctly escaped (a common one is `type`). From the top of my head, these should be things like:

* `import` paths
* `def`, `val` identifiers
* names in argument lists
* others

A possible solution could be to add a new printer combinator that escapes those reserved words and use that instead of raw printing those identifiers.

cb372:
Fixes #114 

Added a new `identifier` combinator to `Printer`, which turns arbitrary strings into valid Scala identifiers by wrapping them with backticks if necessary. Unfortunately it wasn't possible to actually use the combinator in that many places; I ended up directly calling the underlying `Printer.toValidIdentifier` instead.

I didn't want to write my own buggy implementation of the Scala lexical spec, so I used scalameta to take care of the logic to decide when something needs to be backticked.

The printing of Mu schemas/protocols was pretty trivial to fix. The OpenAPI printing, on the other hand, is much more complex. There is a lot more code being generated, and there was already a lot of string manipulation/normalization going on. Because of this, I'm not that confident I've covered every single case correctly, but the test coverage is at least high enough to convince me I haven't made the situation worse.

These changes have caused the OpenAPI codegen behaviour to change in a few cases. Now instead of names being "normalized" (which did not take account of the need to escape Scala reserved words), they are left as-is but wrapped in backticks as necessary when they are used as Scala identifiers. See e.g. [this updated test](https://github.com/higherkindness/skeuomorph/compare/master...cb372:escape-reserved-words?expand=1#diff-199c54d3e60e0f7c8f0c34bd58f7e627R72) for an example of the changes. There were also a couple of small unintended changes that were really hard to avoid without extensive refactoring, e.g. [this change from `Status` to `status`](https://github.com/higherkindness/skeuomorph/compare/master...cb372:escape-reserved-words?expand=1#diff-199c54d3e60e0f7c8f0c34bd58f7e627R390-R391).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -106,7 +106,7 @@ lazy val commonSettings = Seq(
     %%("cats-effect", V.catsEffect),
     %%("circe-core", V.circe),
     %%("circe-parser", V.circe),
-    "org.scalameta"       %% "scalameta"     % "4.3.0",
+    "org.scalameta"       %% "scalameta"     % V.meta,
     %%("cats-laws", V.cats) % Test,
     "io.circe"            %% "circe-testing" % V.circe % Test,
     %%("scalacheck", V.scalacheck)    % Test,
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,7 @@ val V = new {
   val droste           = "0.8.0"
   val kindProjector    = "0.10.3"
   val macroParadise    = "2.1.1"
+  val meta             = "4.3.0"
   val scala212         = "2.12.10"
   val scalacheck       = "1.14.2"
   val specs2           = "4.8.1"
@@ -95,7 +96,6 @@ lazy val commonSettings = Seq(
   crossScalaVersions := Seq(V.scala212),
   ThisBuild / scalacOptions -= "-Xplugin-require:macroparadise",
   libraryDependencies ++= Seq(
-    %%("cats-laws", V.cats) % Test,
     %%("cats-core", V.cats),
     "io.higherkindness"   %% "droste-core"   % V.droste,
     "io.higherkindness"   %% "droste-macros" % V.droste,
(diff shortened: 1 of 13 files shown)


build.sbt
```diff
@@ -95,18 +96,19 @@ lazy val commonSettings = Seq(
...
     "io.higherkindness"   %% "droste-macros" % V.droste,
     "org.apache.avro"     % "avro"           % V.avro,
     "com.github.os72"     % "protoc-jar"     % V.protoc,
     "com.google.protobuf" % "protobuf-java"  % V.protobuf,
     "io.circe"            %% "circe-yaml"    % V.circeYaml,
-    "io.circe"            %% "circe-testing" % V.circe % Test,
     %%("cats-effect", V.catsEffect),
     %%("circe-core", V.circe),
     %%("circe-parser", V.circe),
+    "org.scalameta"       %% "scalameta"     % "4.3.0",
```
juanpedromoreno:
```suggestion
    "org.scalameta"       %% "scalameta"     % V.meta,
```

src/main/scala/higherkindness/skeuomorph/Printer.scala
```diff
@@ -80,6 +82,16 @@ object Printer {
...
 
+  /*
+   * The logic to decide whether a given string needs to be wrapped in backticks
+   * to be a valid Scala identifier is really complicated (the spec is at
+   * https://scala-lang.org/files/archive/spec/2.13/01-lexical-syntax.html#identifiers,
+   * but it's quite wrong/misleading). So we let scalameta take care of it for us.
+   */
+  def toValidIdentifier(string: String): String =
+    if (string.isEmpty) string
+    else scala.meta.Term.Name(string).syntax
```
juanpedromoreno: 👌

pepegar:
Hey @cb372 I'm fairly late to this party... but this looks awesome, thank you very much!

PR #197, opened 2019-12-30
Generate a shapeless Coproduct instead of a placeholder type

cb372:
Previously we were generating a made-up type `Cop[A :: B :: TNil]`. There is [code in the Mu sbt plugin](https://github.com/higherkindness/mu-scala/blob/0dab66791fef6caf7123df613213ba351475049a/modules/idlgen/core/src/main/scala/higherkindness/mu/rpc/idlgen/proto/ProtoSrcGenerator.scala#L65-L71) to rewrite it to `A :+: B :+: CNil` using some regex nastiness. We may as well generate a real type in the first place and remove the need for rewriting.

### DIFF
--- a/src/test/scala/higherkindness/skeuomorph/protobuf/ProtobufProtocolSpec.scala
+++ b/src/test/scala/higherkindness/skeuomorph/protobuf/ProtobufProtocolSpec.scala
@@ -103,7 +103,7 @@ class ProtobufProtocolSpec extends Specification with ScalaCheck {
       |  @_root_.pbdirect.pbIndex(1) name: _root_.java.lang.String,
       |  @_root_.pbdirect.pbIndex(2) books: _root_.scala.Map[_root_.scala.Long, _root_.java.lang.String],
       |  @_root_.pbdirect.pbIndex(3) genres: _root_.scala.List[_root_.scala.Option[_root_.com.acme.book.Genre]],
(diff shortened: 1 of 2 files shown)


src/test/scala/higherkindness/skeuomorph/protobuf/ProtobufProtocolSpec.scala
```diff
@@ -103,7 +103,7 @@ class ProtobufProtocolSpec extends Specification with ScalaCheck {
       |  @_root_.pbdirect.pbIndex(1) name: _root_.java.lang.String,
       |  @_root_.pbdirect.pbIndex(2) books: _root_.scala.Map[_root_.scala.Long, _root_.java.lang.String],
       |  @_root_.pbdirect.pbIndex(3) genres: _root_.scala.List[_root_.scala.Option[_root_.com.acme.book.Genre]],
-      |  @_root_.pbdirect.pbIndex(4,5,6,7) payment_method: Cop[_root_.scala.Long :: _root_.scala.Int :: _root_.java.lang.String :: _root_.com.acme.book.Book :: TNil]
+      |  @_root_.pbdirect.pbIndex(4,5,6,7) payment_method: _root_.scala.Long :+: _root_.scala.Int :+: _root_.java.lang.String :+: _root_.com.acme.book.Book :+: CNil
```
juanpedromoreno: Will we need FQN for `:+:` as well?

cb372: Good point, there's no reason not to. Same for `CNil`. That means we can remove some import-generation logic from the sbt plugin as well.

PR #205, opened 2020-01-03
Use scalameta for codegen

cb372:
I find this easier to work with than the existing printer, which is basically doing fancy string concatenation, because scalameta gives us a nice AST with which to model the code. It also provides quasiquotes, which are very convenient, and it indents the code correctly for us.

(Unfortunately we end up losing a lot of the type information when we do a catamorphism, and we have to add it back in using casting. I haven't found a nice way around this. Open to suggestions.)

The codegen takes a stream type constructor as an argument, allowing the sbt plugin to choose between FS2 and Monix for streaming more cleanly than it does currently. It also includes an `import higherkindess.mu.rpc.protocol._` statement so we don't have to add the import in the sbt plugin.

If people are happy with the scalameta approach, I suggest we remove `higherkindness.skeuomorph.mu.print` and update the Mu sbt plugin to use the new scalameta-based codegen. In future we could also migrate the OpenAPI codegen to scalameta. In this PR I've kept `higherkindness.skeuomorph.mu.print` but added a `@deprecated` annotation.

### DIFF
--- a/docs/docs/docs/index.md
+++ b/docs/docs/docs/index.md
@@ -95,8 +95,8 @@ val avroSchema: Schema = new Schema.Parser().parse(definition)
 val toMuSchema: Schema => Mu[MuF] =
   scheme.hylo(transformAvro[Mu[MuF]].algebra, fromAvro)
 
-val printSchemaAsScala: Mu[MuF] => String =
-  codegen.schema(_).right.get.syntax
+val printSchemaAsScala: Mu[MuF] => Either[String, String] =
+  codegen.schema(_).map(_.syntax)
 
 (toMuSchema >>> println)(avroSchema)
 println("=====")
@@ -152,11 +152,11 @@ val toMuProtocol: Protocol[Mu[ProtobufF]] => mu.Protocol[Mu[MuF]] = { p: Protoco
   mu.Protocol.fromProtobufProto(CompressionType.Identity, true)(p)
 }
 
-val printProtocolAsScala: mu.Protocol[Mu[MuF]] => String = { p =>
+val printProtocolAsScala: mu.Protocol[Mu[MuF]] => Either[String, String] = { p =>
   val streamCtor: (Type, Type) => Type.Apply = {
     case (f, a) => t"_root_.fs2.Stream[$f, $a]"
   }
-  mu.codegen.protocol(p, streamCtor).right.get.syntax
+  mu.codegen.protocol(p, streamCtor).map(_.syntax)
 }
 
 (toMuProtocol >>> println)(protobufProtocol)
--- a/docs/docs/docs/index.md
+++ b/docs/docs/docs/index.md
@@ -42,7 +42,7 @@ libraryDependencies += "io.higherkindness" %% "skeuomorph" % "0.0.19"
 
 ## Examples
 
-### parsing an avro schema and then converting it to scala:
+### Parsing an Avro schema and converting it into Scala code
 
 Given an Avro schema:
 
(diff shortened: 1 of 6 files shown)


src/test/scala/higherkindness/skeuomorph/protobuf/ProtobufProtocolSpec.scala
```diff
@@ -70,6 +72,37 @@ class ProtobufProtocolSpec extends Specification with ScalaCheck {
...
+    }
+
+    val streamCtor: (Type, Type) => Type.Apply = {
+      case (f: Type, a: Type) => t"Stream[$f, $a]"
+    }
+
+    val codegen: higherkindness.skeuomorph.mu.Protocol[Mu[MuF]] => String = {
+      p: higherkindness.skeuomorph.mu.Protocol[Mu[MuF]] =>
+        val tree = higherkindness.skeuomorph.mu.codegen.protocol(p, streamCtor)
+        //println(tree.structure)
```
BenFradet: :art:

src/main/scala/higherkindness/skeuomorph/mu/codegen.scala
```diff
@@ -0,0 +1,170 @@
...
+    // Apply optimizations to normalise the protocol
+    // before converting it to Scala code
+    (nestedOptionInCoproduct[T] andThen knownCoproductTypes[T]).apply(t)
+
+  def protocol[T](
+      protocol: Protocol[T],
+      streamCtor: (Type, Type) => Type.Apply
+  )(implicit T: Basis[MuF, T]): Pkg = {
+    val packageName = protocol.pkg.getOrElse("proto")
+    val packageTerm = packageName.parse[Term].get.as[Term.Ref]
```
BenFradet: I'm not familiar with the scalameta API but is the `get` safe?

cb372: Nope! Need to have a think about error handling. Maybe the whole thing should return an Either. That would also mean we could make the casts slightly more palatable by making them return an Either instead of throwing an exception.

src/test/scala/higherkindness/skeuomorph/protobuf/ProtobufProtocolSpec.scala
```diff
@@ -70,6 +72,37 @@ class ProtobufProtocolSpec extends Specification with ScalaCheck {
...
+        tree.syntax
+    }
+
+    val actual = (parseProtocol andThen codegen)(protobufProtocol)
+
+    val expected = expectation(ct, useIdiom)
+
+    (actual.clean must beEqualTo(expected.clean)) :| s"""
```
cb372: Rather than stripping whitespaces and doing string comparison, we can parse the expected output into a scalameta tree and compare trees instead.

src/main/scala/higherkindness/skeuomorph/mu/codegen.scala
cb372: Done. Everything now returns `Either[String, ?]`.

cb372:
I'm pretty happy with this now. I've requested a review from @pepegar as he wrote the original printer. @juanpedromoreno Do you want to have a look too?

docs/docs/docs/index.md
```diff
@@ -76,26 +76,27 @@ val definition = """
...
+import scala.meta._
 
 val avroSchema: Schema = new Schema.Parser().parse(definition)
 
 val toMuSchema: Schema => Mu[MuF] =
   scheme.hylo(transformAvro[Mu[MuF]].algebra, fromAvro)
 
 val printSchemaAsScala: Mu[MuF] => String =
-  muprint.schema.print _
+  codegen.schema(_).right.get.syntax
```
juanpedromoreno: I know this is an example, but, does it make sense if we show something like this? (taking into account that Either is right-biased):

```scala
codegen.schema(_).map(_.syntax)
```

cb372:
@pepegar I'm going to merge this, but feel free to review later if you have time. I'm happy to make another PR to address any comments.

pepegar:
Sure thing, sorry @cb372 , I reviewed today mornign but didn't submit the approved!  Great job :)

PR #206, opened 2020-01-07
Support protobuf nested messages and enums

cb372:
In `.proto` files it is possible (and quite common) to nest messages and enums inside other messages, and reference them for field types in other message, e.g.

```protobuf
message Person {
  message Name {
    string given_name = 1;
    string family_name = 2;
  }

  Name name = 1;
  int32 date_of_birth = 2;
}

message User {  
  enum Type {
    NORMAL = 0;
    ADMIN = 1;
  }

  int32 id = 1;
  Person.Name name = 2;
  Type type = 3;  
} 
```

This PR adds support for these nested types.

Sorry the diff is pretty huge, but the major changes are:

1. Correctly build the fully-qualified names for nested types when parsing a `.proto` file. (We were generating nonsense names because the logic wasn't written with nested types in mind.)
2. Make sure to collect all messages and enums when parsing a `.proto` file, no matter how deeply nested they are. (We were only looking for messages at one level of nesting, and not picking up nested enums at all.)
3. Add a message's lists of nested messages and enums to `ProtobufF.TMessage`
4. Also add them to `MuF.TProduct` and populate them when transforming from `ProtobufF` to `MuF`
5. Trivially set them to `Nil` when transforming from `AvroF` to `MuF` for now. Avro schemas do support nested records, so we might want to change this in the future.
6. Generate the appropriate scalameta trees for nested types when doing codegen
7. Update the comparison logic to handle nested types, and add a test for it
8. Extend the `ProtobufProtocolSpec` to test the new functionality

The generated code for the above example would look like (slightly simplified):

```scala
case class Person(name: Option[Person.Name], date_of_birth: Int)
object Person {
  case class Name(given_name: String, family_name: String)
}

case class User(id: Int, name: Person.Name, `type`: User.Type)
object User {
  sealed trait Type
  case object NORMAL extends Type
  case object ADMIN extends Type
}
```

### DIFF
--- a/src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
+++ b/src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
@@ -303,7 +303,10 @@ object ParseProto {
     fields.count(f => (f.getNumber == 1 && f.getName == "key") || (f.getNumber == 2 && f.getName == "value")) == 2
 
   def matchNameEntry(name: String, source: DescriptorProto): Boolean =
-    source.getName.toLowerCase == s"${name}Entry".toLowerCase
+    normalizeName(source.getName) == normalizeName(s"${name}Entry")
+
+  def normalizeName(name: String): String =
+    name.toLowerCase.replaceAllLiterally("_", "")
 
   def fromOneofDescriptorsProto[A](
       oneOfFields: List[OneofDescriptorProto],
--- a/src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
+++ b/src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
@@ -182,15 +182,20 @@ object ParseProto {
       protoFields
         .filterNot(f => oneOfNumbers.contains(f.getNumber))
         .map(f => fromFieldDescriptorProto[A](f, descriptor, files))
+    val nestedMessages =
+      descriptor.getNestedTypeList.asScala.toList
+        .filterNot(_.getOptions.getMapEntry) // filter out auto-generated map entry types
+        .map(toMessage[A](_, files))
+    val nestedEnums =
+      descriptor.getEnumTypeList.asScala.toList.map(toEnum[A])
 
-    // TODO need to include nested messages here
     message[A](
       name = descriptor.getName,
       fields = (fields ++ oneOfFields.map(_._1)).sortBy(_.indices.headOption),
       reserved = descriptor.getReservedRangeList.asScala.toList.map(range =>
         (range.getStart until range.getEnd).map(_.toString).toList),
-      nestedMessages = Nil,
-      nestedEnums = Nil
+      nestedMessages = nestedMessages,
+      nestedEnums = nestedEnums
     ).embed
   }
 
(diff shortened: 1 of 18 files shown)


cb372:
@pepegar I agree, and the Cofree idea sounds interesting. I still haven't looked at your UAST stuff properly yet. I think we're in a better position to tackle it now we have a better understanding of the differences between Avro and Protobuf ASTs.

cb372:
@juanpedromoreno I've just tested Mu locally with this change, and the mu-opencensus project is now *so close* to compiling. One compilation error, which I'll investigate tomorrow.

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -152,12 +182,20 @@ object ParseProto {
       protoFields
         .filterNot(f => oneOfNumbers.contains(f.getNumber))
         .map(f => fromFieldDescriptorProto[A](f, descriptor, files))
+    val nestedMessages =
+      descriptor.getNestedTypeList.asScala.toList
+        .filterNot(_.getOptions.getMapEntry) // filter out auto-generated map entry types
```
BenFradet: what does this refer to and why is this necessary?

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -348,26 +371,44 @@ object ParseProto {
...
 
-  def findEnum(name: String, files: List[FileDescriptorProto]): Option[(String, EnumDescriptorProto)] = {
-    case class NamedEnum(fullName: String, enclosingProto: String, enum: EnumDescriptorProto)
-    files
-      .flatMap { f =>
-        val enclosingProto = formatName(f.getName)
-        f.getEnumTypeList.asScala.toList.map(e => NamedEnum(s".${f.getPackage}.${e.getName}", enclosingProto, e))
-      }
+  def findMessage(name: String, files: List[FileDescriptorProto]): Option[NamedDescriptor[DescriptorProto]] =
+    allMessages(files).find(_.fullName == name)
```
BenFradet: :art: it might be a good idea to use `===` just in case this is refactored later

cb372: Good idea

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
cb372: Sorry, the comment is a bit cryptic. I'll expand it with a better explanation.

Maps are encoded in protobuf as lists of (key, value) pairs, with each pair encoded as if it were an embedded message with two fields `key` and `value`. A better explanation is available [in the protobuf docs](https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility).

The protoc tool generates a type for this pair, but we don't care about it (we don't want to generate code for it) so we filter it out.

BenFradet: Thanks a lot!

PR #213, opened 2020-01-14
Support for protobuf signed/unsigned/fixed-width integers

cb372:
* Turn `Muf.TInt` into a coproduct, with separate cases for a "simple" integer and one annotated with Protobuf-specific information.
* To minimise code duplication, I got rid of `TLong`. Gave `TInt` a size, so it can represent both 32-bit and 64-bit integers.
* Fix the Protobuf parser to use the correct integer types
* Update the scalameta codegen to output appropriately tagged types
* Add a test for codegen of all the different Protobuf integer types

As discussed with @pepegar, this design is an attempt to minimise the impact of protobuf-specific "pollution" in the Mu schema. Instead of trying to hide the differences between the feature sets of Protobuf/Avro/OpenAPI/etc and turn MuF into a super-generic schema that can accomodate all of them transparently, we add coproducts with special cases to make it clear exactly where we have special support for Protobuf or other protocol formats.


src/main/scala/higherkindness/skeuomorph/mu/comparison/Comparison.scala
```diff
@@ -182,8 +182,11 @@ object Comparison extends ComparisonInstances {
           val nestedCoproducts = zipLists(path, nc, nc2, _ => Name(n))
           CompareList(fields ++ nestedProducts ++ nestedCoproducts)
 
-        // Numeric widdening
-        case (TInt(), TLong() | TFloat() | TDouble()) | (TLong(), TFloat() | TDouble()) | (TFloat(), TDouble()) =>
+        // Numeric widening
+        // TODO: this is not valid for Protobuf or Avro (e.g. changing an int to a float would be a breaking change)
```
juanpedromoreno: 👍

src/main/scala/higherkindness/skeuomorph/mu/schema.scala
```diff
@@ -28,20 +29,25 @@ import cats.instances.string._
...
-/**
- *
- */
 @deriveTraverse sealed trait MuF[A]
 object MuF {
   @deriveTraverse final case class Field[A](name: String, tpe: A, indices: List[Int])
 
   final case class SumField(name: String, value: Int)
 
+  sealed trait NumberSize
```
juanpedromoreno:
```suggestion
  sealed trait NumberSize extends Product with Serializable
```
?

src/main/scala/higherkindness/skeuomorph/protobuf/schema.scala
```diff
@@ -52,6 +52,11 @@ object FieldF {
   }
 }
 
+sealed trait IntModifier
```
juanpedromoreno:
```suggestion
sealed trait IntModifier extends Product with Serializable
```
?

src/main/scala/higherkindness/skeuomorph/avro/Protocol.scala
```diff
@@ -81,24 +81,24 @@ object Protocol {
...
-    case MuF.TMap(_, value)         => AvroF.map(value)
-    case MuF.TGeneric(_, _)         => ??? // WAT
-    case MuF.TContaining(_)         => ??? // TBD
-    case MuF.TRequired(t)           => T.coalgebra(t)
-    case MuF.TCoproduct(invariants) => AvroF.union(invariants)
-    case MuF.TSum(name, fields)     => AvroF.enum(name, none[String], Nil, none[String], fields.map(_.name))
+    case MuF.TNull()                         => AvroF.`null`()
+    case MuF.TDouble()                       => AvroF.double()
+    case MuF.TFloat()                        => AvroF.float()
+    case x: MuF.TInt[_] if x.size == MuF._32 => AvroF.int()
```
pepegar: I'd rather use structural pattern matching instead.  Othersise we're doing `isInstanceOf` under the hood.

```suggestion
    case MuF.TInt(MuF._32) => AvroF.int()
```

src/main/scala/higherkindness/skeuomorph/avro/Protocol.scala
```diff
@@ -81,24 +81,24 @@ object Protocol {
...
+    case x: MuF.TInt[_] if x.size == MuF._64 => AvroF.long()
```
pepegar: Same here
```suggestion
    case MuF.TInt(MuF._64) => AvroF.long()
```

src/main/scala/higherkindness/skeuomorph/mu/comparison/Comparison.scala
```diff
@@ -182,8 +182,11 @@ object Comparison extends ComparisonInstances {
...
+        case (x: TInt[_], y: TInt[_]) if x.size == _32 && y.size == _64 =>
```
pepegar: Would this work?

```suggestion
        case (TInt(_32), TInt(_64)) =>
```

Probably we need to implement a `TInt.unapply` for it.

src/main/scala/higherkindness/skeuomorph/mu/schema.scala
```diff
@@ -98,24 +109,24 @@ object MuF {
...
-      case TContaining(ts)  => ts.mkString("cont<", ", ", ">")
-      case TEither(l, r)    => s"either<$l, $r>"
-      case TGeneric(g, ps)  => ps.mkString(s"$g<", ", ", ">")
-      case TCoproduct(ts)   => ts.toList.mkString("(", " | ", ")")
-      case TSum(n, vs)      => vs.map(_.name).mkString(s"$n[", ", ", "]")
+      case TNull()                     => "null"
+      case TDouble()                   => "double"
+      case TFloat()                    => "float"
+      case t: TInt[_] if t.size == _32 => "int"
+      case t: TInt[_] if t.size == _64 => "long"
```
pepegar: Same here, If we've the unapply we can have a proper pattern match without guards :D

src/test/scala/higherkindness/skeuomorph/protobuf/ProtobufProtocolSpec.scala
```diff
@@ -331,4 +337,35 @@ class ProtobufProtocolSpec extends Specification with ScalaCheck {
...
+    |    @_root_.pbdirect.pbIndex(4) d: _root_.shapeless.tag.@@[_root_.scala.Int, _root_.pbdirect.Fixed],
+    |    @_root_.pbdirect.pbIndex(5) e: _root_.shapeless.tag.@@[_root_.scala.Int, (_root_.pbdirect.Fixed with _root_.pbdirect.Signed)],
+    |    @_root_.pbdirect.pbIndex(6) f: _root_.scala.Long,
+    |    @_root_.pbdirect.pbIndex(7) g: _root_.shapeless.tag.@@[_root_.scala.Long, _root_.pbdirect.Unsigned],
+    |    @_root_.pbdirect.pbIndex(8) h: _root_.shapeless.tag.@@[_root_.scala.Long, _root_.pbdirect.Signed],
+    |    @_root_.pbdirect.pbIndex(9) i: _root_.shapeless.tag.@@[_root_.scala.Long, _root_.pbdirect.Fixed],
+    |    @_root_.pbdirect.pbIndex(10) j: _root_.shapeless.tag.@@[_root_.scala.Long, (_root_.pbdirect.Fixed with _root_.pbdirect.Signed)]
+    |  )
+    |}
+    |""".stripMargin
```
pepegar: 👏 👏 Awesome to be able to generate this from the pb file, great job!

src/main/scala/higherkindness/skeuomorph/mu/print.scala
```diff
@@ -47,8 +47,7 @@ object print {
       case TNull()                   => "Null"
       case TDouble()                 => "_root_.scala.Double"
       case TFloat()                  => "_root_.scala.Float"
-      case TInt()                    => "_root_.scala.Int"
-      case TLong()                   => "_root_.scala.Long"
+      case _: TInt[_]                => "_root_.scala.Int"
```
BenFradet: I don't think I understand why this is not:

```scala
case x: TInt[_] if x.size == _32 => "_root_.scala.Int"
case x: TInt[_] if x.size == _64 => "_root_.scala.Long"
```

cb372: Because I'm lazy and I don't care about this file anymore 😄 This class is deprecated and I'm planning to make a PR to remove it after the next release. But I'll fix this. Thanks for spotting it.

src/main/scala/higherkindness/skeuomorph/avro/Protocol.scala
cb372: Good point, I thought about adding an unapply but forgot all about it. I'll give it a go.

src/main/scala/higherkindness/skeuomorph/mu/print.scala
BenFradet: right, forgot that this file refers to the old printer

PR #219, opened 2020-01-27
Remove `@message` annotation from generated case classes

cb372:
This annotation was only used for IDL generation, which we don't do anymore.

(also accidentally apply the formatter)

### DIFF
--- a/src/main/scala/higherkindness/skeuomorph/mu/codegen.scala
+++ b/src/main/scala/higherkindness/skeuomorph/mu/codegen.scala
@@ -206,7 +206,7 @@ object codegen {
           nestedCoproducts.traverse(_.as[Stat])
         ).mapN {
           case (args, prods, coprods) =>
-            val caseClass = q"@message final case class ${Type.Name(name)}(..$args)"
+            val caseClass = q"final case class ${Type.Name(name)}(..$args)"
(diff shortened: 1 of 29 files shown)


src/main/scala/higherkindness/skeuomorph/mu/codegen.scala
```diff
@@ -206,7 +206,7 @@ object codegen {
           nestedCoproducts.traverse(_.as[Stat])
         ).mapN {
           case (args, prods, coprods) =>
-            val caseClass = q"@message final case class ${Type.Name(name)}(..$args)"
+            val caseClass = q"final case class ${Type.Name(name)}(..$args)"
```
cb372: Here is the change.

src/test/scala/higherkindness/skeuomorph/protobuf/ProtobufProtocolSpec.scala
```diff
@@ -95,7 +95,7 @@ class ProtobufProtocolSpec extends Specification with ScalaCheck {
       |
       |object book {
       |
-      |@message final case class Book(
+      |final case class Book(
```
cb372: Here is the update to the test.

PR #236, opened 2020-02-21
Cross compilation Scala 2.13

juanpedromoreno:
This PR cross-compile the project with Scala 2.13.


.travis.yml
```diff
@@ -22,23 +23,23 @@ install:
...
+  - sed -i -e 's/TRAVIS_.*_VERSION/^TRAVIS_.*_VERSION=/' .codecov
+  - chmod +x .codecov
+  - sbt ++$TRAVIS_SCALA_VERSION clean test coverageReport coverageAggregate docs/makeMicrosite
+  - if [ $TRAVIS_SCALA_VERSION = "2.13.1" ]; then ./.codecov -X gcov; fi
+
 stages:
-- tests
-- deploy
-- microsite
+  - test
```
cb372: Where are the jobs for this stage defined? I'm also a bit confused by the `script` above. When does that run?

src/main/scala-2.13/higherkindness/skeuomorph/mu/comparison/Comparison.scala
```diff
@@ -0,0 +1,33 @@
...
+trait ComparisonInstances {
+
+  implicit def comparisonCatsFunctor[T] = new Functor[Comparison[T, ?]] {
+    def map[A, B](fa: Comparison[T, A])(f: (A) => B): Comparison[T, B] = fa match {
+      case Comparison.End(res)                  => Comparison.End(res)
+      case Comparison.Compare(a, rep)           => Comparison.Compare(f(a), rep)
+      case Comparison.CompareBoth(x, y)         => Comparison.CompareBoth(f(x), f(y))
+      case Comparison.CompareList(i, rep)       => Comparison.CompareList(i.map(f), rep)
+      case Comparison.MatchInList(a, rep)       => Comparison.MatchInList(a.map(f), rep)
+      case Comparison.AlignUnionMembers(a, rep) => Comparison.AlignUnionMembers(a.view.mapValues(_.map(f)).toMap, rep)
```
cb372: It took me a while to spot the difference 😄

.travis.yml
juanpedromoreno: Without being an expert, I believe that `test` is reserved for the script section. TBH, it's the only way I made it work with the help of @fedefernandez .

src/test/scala-2.12/higherkindness/skeuomorph/openapi/helpers.scala
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package higherkindness.skeuomorph.openapi
+
+trait OperationHelpers {
```
cb372: Rename this file to `OperationHelpers.scala` ?

src/test/scala-2.13/higherkindness/skeuomorph/openapi/helpers.scala
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package higherkindness.skeuomorph.openapi
+
+trait OperationHelpers {
+  import cats.implicits._
+  import schema._
+
+  def operationWithReferences[A](request: Reference, responses: (String, Reference)*): Path.Operation[A] =
+    operationFrom(request.asRight.some, responses.toMap.view.mapValues(_.asRight).toMap)
```
cb372: Did you try scala-collection-compat? I think it will allow us to write `.view.mapValues.toMap` in Scala 2.12, so we don't need to duplicate files.

.travis.yml
fedefernandez: Test is the default one, yes. It's a bit confusing though

juanpedromoreno:
@cb372 Thanks for the suggestion! Ready for another look :)

CC @BenFradet

PR #268, opened 2020-04-26
Update sbt-github to allow overriding organization

alejandrohdezma:
## What has been done in this PR?

- Upgrade `sbt-github` to [latest version](https://github.com/alejandrohdezma/sbt-github/releases/tag/v0.8.1) to allow using `githubOrganization` to override the organization to `47degrees`.
- Move organization settings to the top of `build.sbt` as "build" settings.
- Clean the build a little bit.


build.sbt
```diff
@@ -1,4 +1,6 @@
-import microsites._
+ThisBuild / organization := "io.higherkindness"
+ThisBuild / githubOrganization := "47degrees"
```
pepegar, line 2: Is this correct? shouldn't it be `higherkindness`?

alejandrohdezma: Yep, since we want to override the organization used for docs and POM to be 47 Degrees

Issue #318, opened 2020-08-12
Investigate allowing optional `package` parameters when parsing Protobuf schema


dmarticus:
Per this discussion in gitter
> Anton Semenov @a7emenov 04:40
@dmarticus thank you for the response, really appreciate it! Basically confirms all my assumptions. Though I really love Mu, this may be the reason why we'd have to switch to ScalaPB as it allows to be very flexible when generating packages.
Is it likely that in the future skeuomorph and, by extension, mu will support options like java_package?

It seems like at least _one_ person would want us to allow for optional package generation within skeuomorph and mu.  I'm not sure at this point how significant of an undertaking that work would be but I figured it's worth making a task to track it so we can triage at some point in the future

a7emenov:
@dmarticus, to add some clarity, my problems with the `package` option are:

- It is mandatory for _skeuomorph_ to function, but not for _protoc_.
- It influences where the resulting files will be generated. While there are workarounds around this, I am inclined to believe it should have only "semantic" meaning, i.e. how services should be referenced.

As for a real-world example, my original `.proto` files do not have the `package` option and I cannot really change them as they are used across multiple teams. But even adding the `package` option is kind of fruitless without the `muSrcGenIdiomaticEndpoints` as clients generated by _protoc_ will not be able to reference the service without the package prefix.

dmarticus:
@a7emenov do you have an example protobuf file that i could work with that you'd like to be able to use with mu and skeuomorph?  I'm planning on implementing this behavior this week and I'd love to have a file that you want to be able to convert to use as a test case

a7emenov:
@dmarticus, sure, thank you so much for a swift response.

**Example 1**:

* File:
  ```
  syntax = "proto3";
  option java_package = "my_package";

  service MyService {
    rpc Check(MyRequest) returns (MyResponse);
  }

  message MyRequest {
    string value = 1;
  }

  message MyResponse {
    string value = 1;
  }
  ```

* Expected:
`MyService` service will be generated in "my_package" package.

* Actual:
Service generation fails as the `package` option is not provided.

**Example 2**:

* File:
  ```
  syntax = "proto3";
  package other_package;

  option java_package = "my_package";

  service MyService {
    rpc Check(MyRequest) returns (MyResponse);
  }

  message MyRequest {
    string value = 1;
  }

  message MyResponse {
    string value = 1;
  }
  ```

* Expected:
`MyService` service will be generated in "my_package" package.

* Actual:
[...]

PR #322, opened 2020-08-19
rollback incompatible cross imports

dmarticus:
I accidentally included an import in https://github.com/higherkindness/skeuomorph/pull/321 that doesn't compile correctly in scala 2.12.  My bad.  This change should fix that.

PR #324, opened 2020-08-25
Respect `java_package` option in protobuf files

dmarticus:
Closes https://github.com/higherkindness/skeuomorph/issues/318.  Includes test cases provided by the issue filer :), plus some additional ones that cover missing package info altogether.

This PR is intended to give Skeuomorph comparable behavior to ScalaPB when it comes to handling the package structure -- I want to make it so that the package hierarchy goes like this (1) java package (if present) (2) package name (if present) (3) file name.  More on how ScalaPB does it here [here](https://scalapb.github.io/docs/generated-code/#default-package-structure).

### DIFF
--- a/src/main/scala/higherkindness/skeuomorph/avro/Protocol.scala
+++ b/src/main/scala/higherkindness/skeuomorph/avro/Protocol.scala
@@ -25,7 +25,7 @@ import org.apache.avro.{Schema, Protocol => AvroProtocol}
 import higherkindness.droste._
 import higherkindness.droste.syntax.all._
 
-import scala.jdk.CollectionConverters._
+import scala.collection.JavaConverters._
 
 final case class Protocol[A](
     name: String,
--- a/src/main/scala/higherkindness/skeuomorph/mu/protocol.scala
+++ b/src/main/scala/higherkindness/skeuomorph/mu/protocol.scala
@@ -99,7 +99,7 @@ object Protocol {
     val toImports: DependentImport[T] => DependentImport[U] =
       imp => DependentImport(imp.pkg, imp.protocol, toMu(imp.tpe))
 
-    new Protocol[U](
+    val res = new Protocol[U](
       name = Some(protocol.name),
       pkg = Option(protocol.pkg),
       options = protocol.options,
@@ -116,6 +116,9 @@ object Protocol {
         ),
       imports = protocol.imports.map(toImports)
     )
+
+//    println(s"the resultant protocol is $res")
+    res
 
   }
 
--- a/src/main/scala/higherkindness/skeuomorph/mu/protocol.scala
+++ b/src/main/scala/higherkindness/skeuomorph/mu/protocol.scala
@@ -99,7 +99,7 @@ object Protocol {
     val toImports: DependentImport[T] => DependentImport[U] =
       imp => DependentImport(imp.pkg, imp.protocol, toMu(imp.tpe))
 
-    val res = new Protocol[U](
+    new Protocol[U](
       name = Some(protocol.name),
       pkg = Option(protocol.pkg),
       options = protocol.options,
@@ -116,9 +116,6 @@ object Protocol {
         ),
       imports = protocol.imports.map(toImports)
     )
-
-//    println(s"the resultant protocol is $res")
-    res
 
   }
 
--- a/src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
+++ b/src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
@@ -121,6 +121,38 @@ object ParseProto {
     }
   }
 
+  private def getPackageOrJavaPackage(file: FileDescriptorProto): String = {
+
+    val javaPackage: String = file.getOptions.getJavaPackage
+
+    val filePackage: String = file.getPackage
+
+    val fileName: String = formatName(file.getName)
+
+    if (javaPackage.isEmpty) {
+      if (filePackage.isEmpty)
+        // need to check if this is the behavior we want
+        // or if we should just respect the java package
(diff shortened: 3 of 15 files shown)


src/main/scala/higherkindness/skeuomorph/mu/protocol.scala
```diff
@@ -117,6 +117,9 @@ object Protocol {
       imports = protocol.imports.map(toImports)
     )
 
+//    println(s"the resultant protocol is $res")
+    res
+
```
juanpedromoreno: **suggestion**: Maybe we could revert the changes in this file :)

src/main/scala/higherkindness/skeuomorph/protobuf/schema.scala
```diff
@@ -179,11 +179,12 @@ object ProtobufF {
       def makeOneOfB(oneOf: FieldF.OneOfField[A]): G[FieldF[B]] =
         f(oneOf.tpe).map(b => FieldF.OneOfField[B](oneOf.name, b, oneOf.indices): FieldF[B])
 
-      def traverseFieldF(fieldFList: List[FieldF[A]]): G[List[FieldF[B]]] =
+      def traverseFieldF(fieldFList: List[FieldF[A]]): G[List[FieldF[B]]] = {
         fieldFList.traverse {
           case field: FieldF.Field[A]      => makeFieldB(field).widen
           case oneOf: FieldF.OneOfField[A] => makeOneOfB(oneOf).widen
         }
+      }
```
juanpedromoreno: **question**: why adding braces?

juanpedromoreno (approved):

src/main/scala/higherkindness/skeuomorph/mu/protocol.scala
dmarticus: ooops I thought I did :)

src/main/scala/higherkindness/skeuomorph/protobuf/schema.scala
dmarticus: I _think_ this change is outdated?

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -121,6 +121,34 @@ object ParseProto {
...
+  private def getPackageOrJavaPackage(file: FileDescriptorProto): String = {
+
+    val javaPackage: String = file.getOptions.getJavaPackage
+
+    val filePackage: String = file.getPackage
+
+    val fileName: String = formatName(file.getName)
+    // This package naming behavior is consistent with ScalaPB, see
+    // https://scalapb.github.io/docs/generated-code/#default-package-structure
+    if (javaPackage.isEmpty) {
```
dmarticus, line 133: Quick note here: I implemented this method to stay consistent with the parsing behavior of ScalaPB.  If we want to make it so that we don’t use the filename on the absence of any package information (which would fail the parsing step), I’m open to that as well.  At this point, without better direction, I'm just going for compatibility with an existing solution.

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -121,6 +121,34 @@ object ParseProto {
...
+  private def getTypeNameFromQualifiedName(fullName: String): String =
+    // When comparing the types, we only need to know the name of the type, not the
+    // fully-qualified name (since we use the value of the package passed into the protoc to create
+    // the TNamedType.  However, depending on how we access the type name via `getTypeName`, there is
+    // different behavior depending on whether we access a type name from a file with a `package` header
+    // (which will return the fully-qualified name) vs accessing a type name from a file with a `java_package`
+    // header (which will just return the name of the type).  This method trims the type name from a fully-
+    // qualified name if necessary, which provides consistent behavior when comparing the name of a message
+    // to the name of a type
+    fullName.substring(fullName.lastIndexOf(".") + 1)
```
dmarticus, line 150: Another long comment that basically explains the caveat here: we no longer will be comparing fully-qualified type names to the type that appears in the method, because [getTypeName](https://www.codota.com/code/java/methods/com.google.protobuf.DescriptorProtos$FieldDescriptorProto/getTypeName) behaves differently depending on what it's called against (if called against a file with a `package` header, it returns the fully-qualified type, whereas if called against a file with a `java_package` header, it just returns the typeName).  In this case, we always want to just use the typename, so we parse the string down if necessary.  Yes, I know it's ugly "strings as types" but when working with the java methods that only return strings there's not a lot I can do about it :|

src/test/scala/higherkindness/skeuomorph/protobuf/rating.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package com.acme;
+option java_package = "com.acme";
```
dmarticus, line 3: testing that the `java_package` interop works seamlessly when importing proto files (spoiler: it does)

src/main/scala/higherkindness/skeuomorph/avro/Protocol.scala
```diff
@@ -25,7 +25,7 @@ import org.apache.avro.{Schema, Protocol => AvroProtocol}
 import higherkindness.droste._
 import higherkindness.droste.syntax.all._
 
-import scala.jdk.CollectionConverters._
+import scala.collection.JavaConverters._
```
dmarticus, line 28: fwiw the explanation on why I'm making this change is nicely laid out by Birchall here: https://github.com/higherkindness/skeuomorph/pull/322#issuecomment-682474294

juanpedromoreno (approved):
Great job @dmarticus!

PR #339, opened 2020-09-28
Extend package choosing behavior to protobuf files with `Enum` types

### LINKED ISSUE #337: Invalid enums generated if `package` option is not supplied
After #318 was merged one can successfully generate Scala classes from `.proto` files without having to provide the `package` option. However, this option is still mandatory when generating enum fields in messages as otherwise the resulting field type will be `higherkindness.mu.rpc.protocol.Empty`. 

An example is provided below (note the `enum` field type of the `Example` message):

----
Sample file 1 (invalid generation):
```proto
syntax = "proto3";

message Example {
    ExampleEnum enum = 1;
}

enum ExampleEnum {
    Option1 = 0;
    Option2 = 1;
}
```

Expected result:
```scala
object example {
  final case class Example(@_root_.pbdirect.pbIndex(1) enum: _root_.scala.Option[_root_.example.example.ExampleEnum])
  sealed abstract class ExampleEnum(val value: _root_.scala.Int) extends _root_.enumeratum.values.IntEnumEntry
  object ExampleEnum extends _root_.enumeratum.values.IntEnum[ExampleEnum] {
    case object Option1 extends ExampleEnum(0)
    case object Option2 extends ExampleEnum(1)
    val values = findValues
  }
}
```

Actual result:
```scala
object example {
  final case class Example(@_root_.pbdirect.pbIndex(1) enum: _root_.higherkindness.mu.rpc.protocol.Empty.type)
  sealed abstract class ExampleEnum(val value: _root_.scala.Int) extends _root_.enumeratum.values.IntEnumEntry
  object ExampleEnum extends _root_.enumeratum.values.IntEnum[ExampleEnum] {
    case object Option1 extends ExampleEnum(0)
    case object Option2 extends ExampleEnum(1)
    val values = findValues
  }
}
```

----
Sample file 2 (valid generation):
```proto
syntax = "proto3";

package example;

message Example {
    ExampleEnum enum = 1;
}

enum ExampleEnum {
    Option1 = 0;
    Option2 = 1;
}

```

Result :
```scala
object example {
  final case class Example(@_root_.pbdirect.pbIndex(1) enum: _root_.scala.Option[_root_.example.example.ExampleEnum])
  sealed abstract class ExampleEnum(val value: _root_.scala.Int) extends _root_.enumeratum.values.IntEnumEntry
  object ExampleEnum extends _root_.enumeratum.values.IntEnum[ExampleEnum] {
    case object Option1 extends ExampleEnum(0)
    case object Option2 extends ExampleEnum(1)
    val values = findValues
  }
}
```

dmarticus:
Closes #337.  

cc @a7emenov


src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -146,8 +146,8 @@ object ParseProto {
     // (which will return the fully-qualified name) vs accessing a type name from a file with a `java_package`
     // header (which will just return the name of the type).  This method trims the type name from a fully-
     // qualified name if necessary, which provides consistent behavior when comparing the name of a message
-    // to the name of a type
-    fullName.substring(fullName.lastIndexOf(".") + 1)
+    // to the name of a type.
+    fullName.substring(fullName.indexOf(".", fullName.count(_ == '.')) + 1)
```
BenFradet, line 150: I'm not sure I understand :thinking: , why trim at least the number of dots from the FQN?

Why not just `res0.substring(res0.indexOf('.') + 1)` if you want to keep some of the FQN?

dmarticus: Yeah that's a valid question, let me try to explain as best as I can.

When we create named types for `enum` fields, we have the following code to find the enums in a given protobuf file
```
      case Type.TYPE_ENUM =>
        findEnum(field.getTypeName, files)
          .fold[ProtobufF[Type]](TNull()) { namedEnum =>
            if (makeNamedTypesOptional)
              TOptionalNamedType(namedEnum.scalaPrefix, namedEnum.name)
            else
              TNamedType(namedEnum.scalaPrefix, namedEnum.name)
          }
```

A critical part of this code is `field.getTypeName`, which is an external method provided to us by Google's java protobuf library, and if the protobuf file uses the `package` name at the, as below

```
syntax = "proto3";

package example;

message Example {
    ExampleEnum enum = 1;
}

enum ExampleEnum {
    Option1 = 0;
    Option2 = 1;
}
```

the return value of `getTypeName` will be the fully qualified name _including_ the package name, e.g. `.example.Example.ExampleEnum`

However, if a given protobuf file uses the `java_package` option for specifying the package, such as 
```
syntax = "proto3";

option java_package = "my_package";

service MyService {
  rpc Check(MyRequest) returns (MyResponse);
}

message MyRequest {
  string value = 1;
}

message MyResponse {
  string value = 1;
}
```

the `getTypeName` will only return the FQN with the context of nesting in the file _without_ including the package name (e.g. in this case it would return something like `.MyRequest` instead of `.my_package.MyRequest`.

So, what this method does is return the as much FQN as needed, based on what was passed into it by `getTypeName`; this way, we're able to differentiate between message or enums in a protobuf file that have the same name, but are nested differently.

For example, consider the following messages:

```
    message MessageEvent {
      enum Type {
        TYPE_UNSPECIFIED = 0;
        SENT = 1;
        RECEIVED = 2;
      }
      Type type = 1;
      uint64 id = 2;
      uint64 uncompressed_size = 3;
      uint64 compressed_size = 4;
    }
```
and

```
  message Link {
    bytes trace_id = 1;
    bytes span_id = 2;
    enum Type {
      TYPE_UNSPECIFIED = 0;
      CHILD_LINKED_SPAN = 1;
      PARENT_LINKED_SPAN = 2;
    }
    Type type = 3;
    Attributes attributes = 4;
  }
```

When running this method trimmer on to two different messages, we're able to differentiate between the `.Link.Type` and `.MessageEvent.Type`, so that we don't use the wrong named type when creating the scala file from the protobuf file.

Does that all make sense?  Please feel free to chime in with different ideas about how best to implement this (counting the dots and then indexing based on that seems a _little_ hacky, but it works).  Also, considering this method already has long comment to explain its purpose, I might add some additional context as to why we count the dots.

dzanot (approved):

dmarticus:
@BenFradet let me know what you think of my explanation for your question when you have time :)

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
L-Lavigne: What's the `fullName` we are comparing with in `.find(_.fullName.contains(getTypeNameFromQualifiedName(name)))` ? Is it always a FQDN? If so we wouldn't need to trim anything and could check with `contains(name)` (or preferrably `endsWith`) directly. Or do we have cases where `fullName` is the shorter one and `name` the longer one?

In any case, if we do go with the modified `getTypeNameFromQualifiedName` I suggest renaming it to something like `getUniqueSuffixFromTypeName` or something, and rename the `fullName` argument accordingly.

L-Lavigne (commented):
(See my reply to the comment thread.)

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
dmarticus: Great question!  
 > What's the fullName we are comparing with in .find(_.fullName.contains(getTypeNameFromQualifiedName(name))) ? Is it always a FQN?

 I _think_ it's some form of FQN but the length of it depends on whether we're using `package` version `option java_package`; in the case of `java_package` it's only as long as the nesting of the file that it's in rather than including the package in the name (so it's like an FQN-lite?).  For example, consider these two files:
```
# FILE A
syntax = "proto3";

option java_package = "my_package";

    message MessageEvent {
      enum Type {
        TYPE_UNSPECIFIED = 0;
        SENT = 1;
        RECEIVED = 2;
      }
      Type type = 1;
      uint64 id = 2;
      uint64 uncompressed_size = 3;
      uint64 compressed_size = 4;
    }
```
and 
```
# FILE B
syntax = "proto3";

package my_package;

    message MessageEvent {
      enum Type {
        TYPE_UNSPECIFIED = 0;
        SENT = 1;
        RECEIVED = 2;
      }
      Type type = 1;
      uint64 id = 2;
      uint64 uncompressed_size = 3;
      uint64 compressed_size = 4;
    }
```

calling `getTypeName` the enum for File A will return `.my_package.MessageEvent.Type`, whereas on File B it will just return `.MessageEvent.Type`.  So it's still sorta like an FQN but the second option won't include the package name.

That said, though, we can't use `.endsWith` to compare `fullName` and `name`, because for types such as this example
```
# message A
    message MessageEvent {
      enum Type {
        TYPE_UNSPECIFIED = 0;
        SENT = 1;
        RECEIVED = 2;
      }
      Type type = 1;
      uint64 id = 2;
      uint64 uncompressed_size = 3;
      uint64 compressed_size = 4;
    }
```

and
```
# message B
  message Link {
    bytes trace_id = 1;
    bytes span_id = 2;
    enum Type {
      TYPE_UNSPECIFIED = 0;
      CHILD_LINKED_SPAN = 1;
      PARENT_LINKED_SPAN = 2;
    }
    Type type = 3;
    Attributes attributes = 4;
  }
```

both `fullName`s will end with `Type`, but the `name` value for message A will be `MessageEvent.Type` and the `name` value for message B will be `Link.Type`, so the `.endsWith` value won't be distinct.  

I do agree with your advice on changing the method and parameter names to be more explicit, so thank you for that suggestion!

L-Lavigne: Thanks for the detailed example! It's still a bit unclear to me though - if the `fullName` here can be either `.my_package.MessageEvent.Type` or `.MessageEvent.Type`, then using `endsWith` with no trimming will correctly match `MessageEvent.Type` and not `Link.Type`, in both cases. Could you provide an example that yields a `fullName`/`name` combination where this doesn't hold true?

dmarticus: ohhh I think the problem here is that I didn't understand what `endsWith` would return, I assumed it'd return `Type` but I just read the API doc on `endsWith` and I realized I was WRONG.  I can update this method to be cleaner then; or it's possible I might remove this auxiliary method if it turns out that I can do it in a one-liner.

L-Lavigne: Sounds great! I'm curious to see if the unit tests will turn up some outliers where we'd still need to manipulate the strings. We'll see!

L-Lavigne: (Also just to clarify: `fullName.contains(name)` can still work, `fullName.endsWith(name)` just helps avoid mistakes where `name` would happen to be found somewhere else in `fullName`. My suggestion/question is really about whether `name` needs any manipulation at all to ensure it's a substring of `fullName`, as I haven't seen examples where it does.)

dmarticus: HMMM the plot thickens!  So `.endsWith` works great for `Enum`s but fails for `Message`s (before the same `getUniqueSuffixFromTypeName` method worked for both).  Shoutout unit tests for showing me exactly where that's happening.  Time to go digging!

dzanot (approved):

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -452,9 +459,9 @@ object ParseProto {
 
   def findMessage(name: String, files: List[FileDescriptorProto]): Option[NamedDescriptor[DescriptorProto]] =
     // Note that this method checks that the name of the message matches against the type name that is
-    // extracted from the fully-qualified type name via `getTypeNameFromQualifiedName`.  More context
-    // on why that happens is in the method signature for `getTypeNameFromQualifiedName`.
-    allMessages(files).find(_.fullName.contains(getTypeNameFromQualifiedName(name)))
+    // extracted from the fully-qualified type name via `getUniqueSuffixFromTypeName`.  More context
+    // on why that happens is in the method signature for `getUniqueSuffixFromTypeName`.
+    allMessages(files).find(_.fullName.endsWith(getUniqueSuffixFromTypeName(name)))
```
L-Lavigne, line 464: I see we're still calling the suffix-extraction method - so do we have cases (for messages or enums) where just `fullName.endsWith(name)` doesn't work? I'd be curious to see one.

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -138,16 +138,23 @@ object ParseProto {
...
-    fullName.substring(fullName.lastIndexOf(".") + 1)
+    // to the name of a type.
+    //
+    // In addition, the reason we count '.' characters is because the value of the FQN will be different
+    // depending on whether the fullName was generated from a protobuf file that used the `package` option,
+    // vs a protobuf file that used the `java_package`, or simply had an empty option.  Therefore, we need
+    // to parse exactly as many dots as the FQN so that we extract exactly as much FQN as we need to use
+    // the type name appropriately.  A more involved example explaining why we do this can be found here:
+    // https://github.com/higherkindness/skeuomorph/pull/339#discussion_r496969857
+    typeName.substring(typeName.indexOf(".", typeName.count(_ == '.')) + 1)
```
L-Lavigne, line 157: I didn't comment on this earlier in case we ended up removing this method - but what is the purpose of `indexOf(".", typeName.count(_ == '.')`? It looks like the intention is to match the n'th occurence of `"."` in `typeName`, but actually the second parameter of `indexOf` is just the substring index to start the search at.

dmarticus: thanks for taking the time to dive into this with me!  I honestly don't know why I thought that counting the `.`s was an effective strategy; I feel like matching against `.endsWith` has to be the way to go I just couldn't get it together.  I'm gonna try again today now that I have a little bandwidht.

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
dmarticus: Here's an example case where the `fullName.endsWith(name)` doesn't work

```
syntax = "proto3";

package other_package;

option java_package = "my_package";

service MyService {
  rpc Check(MyRequest) returns (MyResponse);
}

message MyRequest {
  string value = 1;
}

message MyResponse {
  string value = 1;
}
```

In the case for this protobuf file, we want the FQN for the `MyRequest` message to look like `$fileName.my_package.MyRequest`, because the java_package name trumps in priority per the protobuf docs.  

However, when we call `findMessage(field.getTypeName, files)`, two things happen:

1. the FQN returned for the `MyRequest` message from `field.getTypeName` is `other_package.MyRequest` (where we'd hope it to be `my_package.MyRequest`, this is what I'm trying to work around(
2. `findMessage` fails to find anything because when it tries to do `fullName.endsWith(name)`, `fullName == my_package.MyRequest` and `name == other_package.MyRequest`, so the `.endsWith` returns false and we're sad.

So, what I'm hoping to do with my `getUniqueSuffixFromTypeName` method is to extract the minimum meaningful suffix from the typeName returned from `field.getTypeName` that we need to match the types, and _this_ is why a simple `.endsWith` doesn't work for me.

I'd love to have some other ideas on how to best work around this because otherwise I can't unsee my existing implementation, lol.

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
dmarticus: OKAY so it turns out it's still complicated in but I wrote up a specific example of why in the other comment so check it out @L-Lavigne :)

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -476,12 +467,9 @@ object ParseProto {
       e <- m.desc.getEnumTypeList.asScala
     } yield {
       val parentMessageNames = m.parentMessageNames :+ m.name
-      NamedDescriptor(m.pkg, m.enclosingProto, parentMessageNames, e.getName, e)
+      NamedDescriptor(m.fqn, m.packageName, m.javaPackage, m.enclosingProto, parentMessageNames, e.getName, e)
```
L-Lavigne, line 470: This could be simplified to `yield m.copy(parentMessageNames = m.parentMessageNames :+ m.name, desc = e)`

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -47,14 +47,16 @@ object ParseProto {
   )
 
   final case class NamedDescriptor[A](
-      pkg: String,
+      fqn: String,
+      packageName: String,
```
L-Lavigne, line 51:
```suggestion
      scalaPackage: String,
      filePackage: String,
```

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -450,17 +443,22 @@ object ParseProto {
     }
   }
 
+  private def matchOnPackageParameters[A](name: String, file: NamedDescriptor[A]) = {
```
L-Lavigne, line 446: I suggest integrating this method into the initializer of `NamedDescriptor.fullName`  (which isn't directly used anywhere else) and then simply calling `file.fullName.endsWith(name)`. 

Perhaps we could also change `fullName` to a `val` and rename it to something more descriptive but I don't have any idea for it at the moment.

L-Lavigne: Following up on that - we could have the following implementations in `NamedDescriptor`:
```
    val fullName: String =
      (filePackage.split('.').toList ++ parentMessageNames ++ List(name)).mkString(".", ".", "")

    val scalaPrefix: List[String] =
      scalaPackage.split('.').toList ++ List(enclosingProto) ++ parentMessageNames
```

I don't know if it would work for all cases, but it passes the unit tests at least. Also it means we could remove the `javaPackage` parameter and `fqnParts` member.

dmarticus: I like this idea a lot!  Thanks for sharing!  I will take a crack at this implementation :)

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -121,7 +123,7 @@ object ParseProto {
     }
   }
 
-  private def getPackageOrJavaPackage(file: FileDescriptorProto): String = {
+  private def namePackage(file: FileDescriptorProto): String = {
```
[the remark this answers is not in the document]
L-Lavigne: I think we should call it `scalaPackage` since that's what the ScalaPB doc (linked in the method's comment) says it is. We'd also rename the `NamedDescriptor` parameter accordingly. (As per my other suggestion: `fqn` would become `scalaPackage` and `packageName` would become `filePackage`).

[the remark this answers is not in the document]
dmarticus: I addressed this comment and your other one in my most recent change!  @L-Lavigne :)

dmarticus:
@L-Lavigne any insights as to why this dropped coverage?  I added more tests than before but the code coverage bot says I decreased coverage overall... might add some additional unit tests to see if that helps.

src/main/scala/higherkindness/skeuomorph/protobuf/ParseProto.scala
```diff
@@ -66,15 +66,26 @@ object ParseProto {
...
+
+    /*
+     * This is the name of the file that we use when parsing types from the actual protobuf file.
+     * Depending on whether the file has a `java_package` or not, this name will either use the
+     * `package` or the `java_package` in order to match the types correctly (`field.getTypeName`
+     * returns inconsistent values for the package name when called on a protobuf file that has both
+     * a `java_package` and a `package`, so this approach makes sure that we always match with the correct
+     * value).
+     */
+    val fullName: String =
```
L-Lavigne, line 87: I don't think we need this one anymore, in my experiments I was matching on `fullProtoName` instead (since that one is specially built using the `filePackage` and not the `javaPackage`). If that works we can eliminate the `javaPackage` constructor parameter altogether.

dmarticus: oh yeah good call!  Sorry; thank you for clarifying :)

L-Lavigne (approved):
Great job! 🎉

PR #340, opened 2020-09-30
Adds logical and fixed types to code generation

dzanot:
This is the first step to replacing avro hugger code generation with skeuomorph codegen in `sbt-mu-srcgen` outlined in [this ticket](https://github.com/47deg/marlow/issues/543). 
It adds the logical types that are supported in the `1.10.0` avro compiler, as well as modifying the `MuF.TByteArray` type to include the info of the avro `fixed` types.


build.sbt
```diff
@@ -52,26 +52,25 @@ lazy val documentation = project
...
+    "io.higherkindness"  %% "droste-core"       % "0.8.0",
+    "io.higherkindness"  %% "droste-macros"     % "0.8.0",
+    "org.apache.avro"     % "avro"              % "1.10.0",
+    "com.github.os72"     % "protoc-jar"        % "3.11.4",
+    "com.google.protobuf" % "protobuf-java"     % "3.12.2",
+    "io.circe"           %% "circe-core"        % "0.13.0",
+    "io.circe"           %% "circe-parser"      % "0.13.0",
+    "io.circe"           %% "circe-yaml"        % "0.13.1",
+    "org.scalameta"      %% "scalameta"         % "4.3.22",
+    "com.julianpeeters"  %% "avrohugger-core"   % "1.0.0-RC22" % Test,
```
dzanot, line 66: I removed `avro-compiler` in favor of `avrohugger-core` as the latter with remove imported structures from the resulting `org.apache.avro.Protocol` structure. 
This caused some incompatibility issue with the `scala-collection-compat` library so I removed that dependency as well which didn't cause any apparent failures.

src/main/scala/higherkindness/skeuomorph/mu/protocol.scala
```diff
@@ -65,21 +65,26 @@ object Protocol {
...
           response = OperationType(toMu(msg.response), false)
         )
-
-    Protocol(
-      name = None,
-      pkg = proto.namespace,
-      options = Nil,
-      declarations = proto.types.map(toMu),
-      services = List(
+    val services = if (proto.messages.isEmpty) {
```
dzanot, line 68: This is to prevent an empty service when there are no rpc methods defined in the avdl

src/main/scala/higherkindness/skeuomorph/mu/codegen.scala
```diff
@@ -154,15 +155,22 @@ object codegen {
...
+      case TNull()                      => t"_root_.higherkindness.mu.rpc.protocol.Empty.type".asRight
+      case TDouble()                    => t"_root_.scala.Double".asRight
+      case TFloat()                     => t"_root_.scala.Float".asRight
+      case x @ TInt(_)                  => intType(x).asRight
+      case TBoolean()                   => t"_root_.scala.Boolean".asRight
+      case TString()                    => t"_root_.java.lang.String".asRight
+      case TByteArray(Length.Arbitrary) => t"_root_.scala.Array[Byte]".asRight
+      case TByteArray(Length.Fixed(n, _, l)) =>
+        val aliasedType = t"_root_.scala.Array[Byte]"
+        q"""object ${Term.Name(n)} {
```
dzanot, line 167: Rendering to just a type alias of a byte array for now. I wanted to tag it with a literal type but because we are cross building to 2.12 it wasn't going to be super clean

src/test/scala/higherkindness/skeuomorph/avro/AvroSpec.scala
```diff
@@ -77,9 +82,27 @@ class AvroSpec extends Specification with ScalaCheck {
...
+      "LogicalTypes",
+      "NestedRecords",
+      "ImportedService",
+      "Fixed"
+    ).map(checkGenerated)
+
+  def checkGenerated(idlName: String) = {
+    val idlResourceName = s"avro/${idlName}.avdl"
+    val idlUri = Try(getClass.getClassLoader.getResource(idlResourceName).toURI)
+      .getOrElse(throw new Exception(s"No avdl found for name: ${idlName}"))
```
L-Lavigne, line 97: Just a nitpick for a few places in this file: if I don't have a specific exception to throw I like to shorten `throw new Exception` to `sys.error`. Also some of the interpolated strings here don't need curly brackets.

src/test/scala/higherkindness/skeuomorph/avro/AvroSpec.scala
```diff
@@ -77,9 +82,27 @@ class AvroSpec extends Specification with ScalaCheck {
...
+    val idlFile = new File(idlUri)
+    val avroProtos =
+      (new FileInputParser).getSchemaOrProtocols(idlFile, Standard, new ClassStore, getClass.getClassLoader).collect {
+        case Right(protocol) => protocol
+      }
+    val avroProto = avroProtos
+      .find(p => p.getName == idlName)
```
L-Lavigne, line 104: Another nitpick: we can probably combine this check with the `collect` partial function above (changing it to `collectFirst`).

src/test/scala/higherkindness/skeuomorph/avro/AvroSpec.scala
```diff
@@ -89,53 +112,41 @@ class AvroSpec extends Specification with ScalaCheck {
       t"Stream[$f, $a]"
     }
 
-    val actual = codegen.protocol(muProto, streamCtor).right.get
+    val actual = codegen.protocol(muProto, streamCtor).fold(s => throw new Exception(s), identity)
```
L-Lavigne, line 115: Perhaps we could wrap `expected` in an Either `Right` so we don't need to extract `actual` before comparing them.

L-Lavigne (dismissed):
Good work! Approved with some nitpicks in one of the tests.

src/test/scala/higherkindness/skeuomorph/avro/AvroSpec.scala
dzanot: Changed both of these

dzanot:
@L-Lavigne I've addressed your comments, and made some changes since merging latest master. Can you take another look?

src/test/scala/higherkindness/skeuomorph/avro/AvroSpec.scala
L-Lavigne: @dzanot I see this wasn't applied so just curious, would there be issues with such a change? Perhaps it would hinder the custom comparison message in which case we'd probably have to keep the test as-is.

dzanot: This actually changed but the test is in the new `AvroProtocolSpec.scala` file here https://github.com/higherkindness/skeuomorph/pull/340/files#diff-791c60e0123c350ef43edff0c1b159887e0163be800c85b7b700494cab9edfa7R117
Not sure why github is still showing this one though

L-Lavigne: Ah OK thanks. Looks like we're still doing `right.get` on `actual` though: https://github.com/higherkindness/skeuomorph/pull/340/files#diff-791c60e0123c350ef43edff0c1b159887e0163be800c85b7b700494cab9edfa7R69

dzanot: Ah yeah I think this is because we are calling `.isEqual` for comparison. I tried refactoring so both actual and expected were `Either`, but using the specs2 `shouldEqual` matcher is using different equality. I can keep poking to see if I can configure specs2 it to use the scala meta `isEqual`

L-Lavigne: Ah no worries then, it would be a nice to have but not worth spending too much time on IMO.

L-Lavigne (approved):

dzanot:
Sorry @L-Lavigne ... I removed some unused imports and dismissed your approval

L-Lavigne (approved):
@dzanot Could you add some labels to the PR for the auto-changelog?

PR #342, opened 2020-10-02
Stop setting `methodNameStyle = Capitalize` when useIdiomaticEndpoints is true

L-Lavigne:
This PR addresses [sbt-mu-srcgen #93](https://github.com/higherkindness/sbt-mu-srcgen/issues/93) along with an upcoming code-change PR for `sbt-mu-srcgen` and a documentation update PR for `mu-scala`.

The main functional change is that we no longer set `methodNameStyle = Capitalize` when `useIdiomaticEndpoints` is true. The flag gets the default value of `true` for clarity.

~It does so by splitting the `useIdiomaticEndpoints` source-generation flag (created in [mu-scala #599](https://github.com/higherkindness/mu-scala/issues/599) to control service method capitalization and package prefixes), into 2 distinct flags with different defaults:~

~1) `useIdiomaticGrpc` which controls the namespace prefix and defaults to `true`;~
~2) `useIdiomaticScala` which controls the endpoint capitalization and defaults to `false`.~

~More formally, if `useIdiomaticScala` is true and if _all_ the RPC call definitions in an input IDL are capitalized, the Scala methods in the services generated from this IDL will be _de_-capitalized, while the gRPC endpoints derived from those will be re-capitalized to match the original IDL. For example, a Protobuf IDL `rpc GetBook` definition becomes a Scala `def getBook` method inside a `methodNameStyle = Capitalized` service, which then generates a `GetBook` gRPC endpoint. A `getBook` IDL definition (for example from Avro) would remain unchanged (`methodNameStyle = Unchanged`, `def getBook`, and `getBook` gRPC endpoint).~

~The reason for the limitation that all RPC calls must be capitalized for the option to apply, is that we do not annotate individual methods, only the top-level `@service` annotation. Thus if we were to de-capitalize a mixed-capitalization group of calls and apply `methodNameStyle = Capitalized` to the service, some of the originally-lowercased methods would be incorrectly re-capitalized which is part of the issue we're addressing here (for example, `getBook` and `GetBook` would both become `def getBook` Scala and `GetBook` endpoints).~

Note that the PR also updates the documentation, moves some test resources into the `resource` root, splits and renames tests between Schema, Protocol, and Cats laws classes, and normalizes the tests a bit between the Avro and Proto cases.

I'm happy to provide some more background on the "idiomatic" flag's history as I understand it, in case it helps in reviewing this change. Feel free to ask about this or anything else unclear.


src/main/scala/higherkindness/skeuomorph/StringUtils.scala
```diff
@@ -0,0 +1,5 @@
+package higherkindness.skeuomorph
+
+object StringUtils {
+  def decapitalize(s: String): String = if (s.isEmpty) s else s"${s.head.toLower}${s.tail}"
```
juanpedromoreno, line 4: Could we use `s.capitalize`?

src/main/scala/higherkindness/skeuomorph/avro/schema.scala
```diff
@@ -29,7 +29,7 @@ import org.apache.avro.Schema
 import org.apache.avro.Schema.Type
 import higherkindness.droste.{Algebra, Coalgebra}
 
-import scala.collection.JavaConverters._
+import scala.jdk.CollectionConverters._
```
juanpedromoreno, line 32: I'm invoking @dmarticus here, as he has been experiencing some issues lately. I'm not sure if this causing binary issues in the sbt plugin side.

cb372:
Maybe a silly question, but do we really need these flags?

I don't see any reason to ever disable `useIdiomaticGrpc`. As I understand it, including the namespace in the service name is not just "idiomatic", it's de-facto standard. I'd say the fact that Mu didn't do this originally was probably a bug.

>if useIdiomaticScala is true and if all the RPC call definitions in an input IDL are capitalized, the Scala methods in the services generated from this IDL will be de-capitalized, while the gRPC endpoints derived from those will be re-capitalized to match the original IDL

This seems like an awful lot of work just to avoid a capital letter at the start of a method name. Do we have any evidence that our users actually care about this?

Maybe we can reduce complexity by being a bit more opinionated and less configurable?

src/main/scala/higherkindness/skeuomorph/avro/schema.scala
dmarticus: What up!  I've been invoked!  

Anyway, JP is right in that making this change will cause binary compat issues on the sbt plugin (specifically the `sbt-mu-srcgen` plugin) side, which is super obnoxious but unfortunately is something we'll have to live with until Davis finishes the work migrating away from avrohugger.  Chris sums it the reasons _why_ we have bincompat issues excellently in [this comment](https://github.com/higherkindness/skeuomorph/pull/322#issuecomment-682474294), but a quick TL;DR is that avrohugger defines its own scala.jdk.CollectionConverters object, which means the object with the same name provided by scala-collection-compat is invisible.

In the short term, this means that we can't get rid of these 2.13 warnings until we migrate off avrohugger, but since both Lawrence and I have tried in different PRs to skeuomorph, I'm gonna write a quick issue explaining this problem and then tie that issue to the completion of the avrohugger migration task so we can make sure that we address this obnoxious warning as soon as we can.

dmarticus: issue: https://github.com/higherkindness/skeuomorph/issues/343

L-Lavigne: Thanks @dmarticus for the explanation. Not sure why AvroHugger decided to clash that namespace, but in any case I've reverted the Converters change for now.

L-Lavigne:
> Maybe a silly question, but do we really need these flags?

@cb372 After some reflection I agree about both flags. For `useIdiomaticGrpc`, keeping it around could help legacy services that may have grown dependent on the non-standard gRPC endpoints when calling them from outside Mu, but that should be a rare case and anyway we decided to allow a breaking change here.

As for `useIdiomaticScala` and the `methodNameStyle` flag it controls, they're mostly holdovers from when we had an IDL-generation-from-source feature and we often had idiomatic Scala methods as a source and idiomatic Protobuf as a destination. (See [mu-scala #599](https://github.com/higherkindness/mu-scala/issues/599).) It's a bit more awkward to implement in reverse as we're seeing, and perhaps a user insisting on idiomatic method names could just add a layer of indirection to the calls. If we do get rid of that flag we may as well get rid of the `methodNameStyle` annotation parameter as well since there's no longer a use case where it would have a value of `Capitalize`.

Thoughts?

cb372:
So to summarise, 

* keep `useIdiomaticGrpc`, only for the sake of legacy users who might need it
* lose `useIdiomaticScala`
* but keep the underlying `methodNameStyle` flag in Mu, for people who use Mu but don't use sbt-mu-srcgen and might want the method names capitalised in gRPC

Sounds good to me, assuming my understanding is correct.

src/main/scala/higherkindness/skeuomorph/avro/schema.scala
dzanot: This might still be blocked after my task is done. I'm still planning on using avrohugger to generate the `org.apache.avro.Protocol` objects while then piping the resulting `Protocols` into skeuomorph.  The reason for this is because avrohugger does stuff to strip out imported protocols from the final `Protocol`, something the native `org.apach.avro.idl.Idl` parser does not do which results in a giant protocol containing all imported definitions.

dzanot: I've found that adding avrohugger as a test dependency to skeuomorph is actually causing bin incompat issues in the tests... so back to the drawing board 😩

L-Lavigne:
@cb372 I removed `useIdiomaticScala` and kept the name `useIdiomaticEndpoints` for the gRPC package prefixes for maximum compatibility, and made it `true` by default for clarity. Please let me know if that makes sense.

cb372 (approved):
LGTM. Thanks for all the miscellaneous cleanup as well.

PR #472, opened 2021-06-11
Update to cats-effect 3

Daenyth:


juanpedromoreno (commented):
Probably you are only missing `import cats.effect.unsafe.implicits.global` on the docs. Other than that LGTM! Thanks @Daenyth !

juanpedromoreno (approved):
Great, thanks @Daenyth !

